PR #4, opened 2020-10-15
Home page UI

rohithdara:
All buttons create an alert that says what the button's function should be. The one thing I couldn't figure out is getting the button functionality to work across the entire button for the start cycle button. It only works in the same horizontal plane where the text exists.


components/home/Home.jsx
```diff
@@ -1,10 +1,85 @@
...
+
+export default class App extends React.Component {
+
+   render() {
+      return (
+         <View style = {styles.container}>
+
+
+            {/*Cycles Section*/}
+            <View style = {[{width: "95%", height: "40%", margin: 30, backgroundColor: "F6F6F6", borderWidth: 5, borderRadius: 15, borderColor: "#5DB075"}]}>
```
jpoist97: Which way of applying style to elements do you prefer? I think we should pick one so all of our code is the same. I personally prefer defining them in the styles object you have at the bottom, I think it makes the code more readable.

components/home/Home.jsx
```diff
@@ -1,10 +1,85 @@
...
+            </View>
+
+
+            {/*Workouts Section*/}
+            <View style = {[{width: "95%", height: "40%", margin: 10, backgroundColor: "#F6F6F6", borderWidth: 5, borderRadius: 15, borderColor: "#5DB075"}]}>
+               <Text style = {styles.title}>
+                  Workouts
+               </Text>
+               <Text style = {[{position: "absolute", top: 82, left: 10, fontSize: 18}]}>
+                  Workout 1
```
jpoist97: This will be a variable name that we get from our database right? you could represent this with like test data for now (declare a variable at the top of your code and fill it temporarily). Then you can just render the variable here instead of "workout 1"

jpoist97: Also, are we always going to display just two workouts here? What if the user only has created one workout? We might want to make this part dependent on a list of workout objects

jpoist97 (commented):
Looks really good overall 👍 , I made a couple small comments but nothing too big. One thing is that whichever way you decide to do style just do it for the whole component so its consistent.

components/home/Home.jsx
rohithdara: I agree if we can figure out how to separate the shared style attributes and the unique attributes.

components/home/Home.jsx
rohithdara: Yes, it will later be replaced with a database query. As per your suggestion, I declared 2 variables at the top for this temporarily

rohithdara: Yeah my idea was that for now, we just always display 2. However, when we do have a list of workouts per user, it should be different based on how many workouts exist. I think in the workouts tab, if they have more than 2 workouts, there should be an option to change which ones are on the home page. If there is only 2 workouts both show on the home page. If there is only 1 workout, it will show the first one and the second one is replaced with "Make another workout in the workouts tab to show up here" or something.

components/home/Home.jsx
rohithdara: Fixed

components/home/Home.jsx
```diff
@@ -1,10 +1,101 @@
...
+
+
+            {/*Cycles Section*/}
+            <View style = {{...styles.sectionComponent, margin: 30}}>
+
+               <Title style = {styles.title}>
+                  Cycles
+               </Title>
+
+               <Button contentStyle={{width: 331, height: 170}} style = {[{backgroundColor: "#5DB075", borderRadius: 20, borderColor: "#5DB075", position: "absolute", top: 60, alignSelf: "center"}]} color = "#FFFFFF" onPress = {() => alert("This should start the current workout in the cycle")}>
```
jpoist97: nitpick: move these some of these styles to stylesheet

components/home/Home.jsx
```diff
@@ -1,10 +1,101 @@
...
+                  Workouts
+               </Title>
+
+               <Subheading style = {[{position: "absolute", top: 82, left: 10, fontSize: 18}]}>
+                  Workout 1
+               </Subheading>
+
+               <Subheading style = {[{position: "absolute", top: 157, left: 10, fontSize: 18}]}>
+                  Workout 2
+               </Subheading>
```
jpoist97: So you said that these workout will be dynamic based off a list. Try doing the same thing you did with the variables at the top with this section but use a list instead in case we have 0 or 1 workouts.

jpoist97 (commented):
Almost done, lmk if you're confused about the list thing or if you don't think we need it

finlaylp (dismissed):
Looks good.

components/home/WorkoutContainer.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+    {props.items.map((item, index) => (
+      <Subheading style={[{
+        position: 'absolute', top: textStartLocation + index * 70, left: 15, fontSize: 18,
+      }]}
+      >
+        {item.name}
+      </Subheading>
+    ))}
+  </View>
```
jpoist97, line 25: You only need one map for this part, you can return both of the tags in the function.

components/home/WorkoutContainer.jsx
```diff
@@ -0,0 +1,35 @@
...
+  <View>
+    {props.items.map((item, index) => (
+      <Button contentStyle={{ width: 100, height: 35 }} style={{ ...styles.startButton, top: buttonStartLocation + index * 70 }} color="#FFFFFF" onPress={() => alert(`This should start recording ${item.name}`)}>
+        Start
+      </Button>
+    ))}
+
+    {props.items.map((item, index) => (
+      <Subheading style={[{
+        position: 'absolute', top: textStartLocation + index * 70, left: 15, fontSize: 18,
```
jpoist97: Also is there a reason you chose to absolutely position these elements, I think it makes more sense to have them be relatively positioned but if this works that's just a personal preference kinda thing

rohithdara: Personal preference for it I guess for me too. If it becomes an issue when I implement the functionality, I'll change it to relative positioning

jpoist97 (approved):

PR #5, opened 2020-10-16
Created Profile Page UI, added react-native-paper to project, Created theme

ctperry0301:
Screen Shot 2020-10-15 at 5 22 59 PM


components/profile/Profile.jsx
```diff
@@ -1,11 +1,73 @@
...
 
 export default (props) => {
+  const [name, setName] = React.useState('John Doe');
+  const [state, setState] = React.useState('6x a week');
    return (    
-      <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
-         <Text>Profile Screen</Text>
-         <Text>Replace these Text tags with your component</Text>
+    <SafeAreaView style={{containier:  {flex: 1}}}>
+      <View style={{flexDirection: 'column', alignItems: "stretch" }}>
```
jpoist97: Looks good, one small thing is do you want to have the style be in a styles object at the bottom of the component and then we kind of use that as our stylesheets, it lets us do it like classes and I think its a bit cleaner and easier to read

jpoist97 (dismissed):

components/profile/Profile.jsx
```diff
@@ -1,10 +1,100 @@
...
+  return (
+    <SafeAreaView style={{ containier: { flex: 1 } }}>
+      <View style={{ flexDirection: 'column', alignItems: 'stretch' }}>
+        {/* Start ofRow containing, settings, profile, and logout */}
+        <View style={{ flexDirection: 'row', justifyContent: 'center', height: '8%' }}>
+          <View style={{ width: '25%' }}>
+            <Button
+              uppercase={false}
+              compact
+              onPress={() => Alert.alert('Settings Pressed')}
```
finlaylp, line 22: Is this going to stay anon here? Or will there be a func defined later?

finlaylp (approved):
Should probably make the goals part in profile dynamic soon. I didn't fully understand everything, still new to this, but might be worth defining functions for the onPress. Otherwise looks good for now

PR #9, opened 2020-10-26
Dev 1 signup

finlaylp:
I'll move it into the authentication folder from login branch once it's approved.


finlaylp:
Idk what happened with package lock json again

signup.js
```diff
@@ -0,0 +1,194 @@
+import React from 'react';
```
jpoist97: this should be a .jsx file, also move this to the components folder and just put it in the authentication folder, you won't have merge conflicts

jpoist97: this matters bc the linting only runs on jsx files in components folder

signup.js
```diff
@@ -0,0 +1,194 @@
...
+        //Decdie on medium font
+      },
+      light: {
+        //Decide on light font
+      },
+  
+    },
+  };
+  
+  const theme = {
```
jpoist97: I haven't looked at theme yet, could you just briefly tell me what you're using it for here. Also @ctperry0301 is this something we only want once in our app since we have it currently in App.js

jpoist97 (commented):
go ahead and move this file into components and run the linting, then request my review again. Also same thing here with the package-lock, when I run `npm i` it deletes the lines it says you added to it. Try running that and see if it does it for you too. I remember you said you modified smt in node_modules so if I'm remembering correctly that might be why

signup.js
ctperry0301: Yeah. Everything related to Theme should only be in the App.js file. The theme will be applied everywhere across the site that is nested within <Provider theme={theme}> in the App.js file.

finlaylp: Only had that there temporarily so that I could see my code, I set it as the appentry and needed it to be inside the theme. Taking it out now it shouldn't have been included in the PR

ctperry0301 (approved):
Looks Good to me. In the future, try to add a screenshot of the changes that you made to the github PR. (Its not that important right now cause of the styling overview)

PR #10, opened 2020-10-30
[DEV-5] Implement Cycle View Page

jpoist97:
Implemented Cycle view page. Added Montseratt fonts. Added styled-components as an alternative to stylesheets. Want it for color based on props here and for styling based on props in log workout component.

IMG_3727

### DIFF
--- a/App.js
+++ b/App.js
@@ -9,8 +9,31 @@ import Calendar from './components/calendar/Calendar';
 import Home from './components/home/Home';
 import Profile from './components/profile/Profile';
 import Workouts from './components/workouts/Workouts';
+import Cycles from './components/cycles/Cycles';
 import { configureFonts, DefaultTheme, Provider as PaperProvider } from 'react-native-paper';
-
+import { Entypo } from '@expo/vector-icons';
+import { AppLoading } from 'expo';
+import {
+  useFonts,
+  Montserrat_100Thin,
+  Montserrat_100Thin_Italic,
+  Montserrat_200ExtraLight,
+  Montserrat_200ExtraLight_Italic,
+  Montserrat_300Light,
+  Montserrat_300Light_Italic,
+  Montserrat_400Regular,
+  Montserrat_400Regular_Italic,
+  Montserrat_500Medium,
+  Montserrat_500Medium_Italic,
+  Montserrat_600SemiBold,
+  Montserrat_600SemiBold_Italic,
+  Montserrat_700Bold,
+  Montserrat_700Bold_Italic,
+  Montserrat_800ExtraBold,
+  Montserrat_800ExtraBold_Italic,
+  Montserrat_900Black,
+  Montserrat_900Black_Italic,
+} from '@expo-google-fonts/montserrat';
 
 const Tab = createBottomTabNavigator();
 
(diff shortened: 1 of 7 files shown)


components/cycles/Cycles.jsx
```diff
@@ -0,0 +1,53 @@
+import * as React from 'react';
+import { SafeAreaView, TouchableOpacity, Text } from 'react-native';
+import styled from 'styled-components/native';
+import { Entypo } from '@expo/vector-icons';
+import AllCycles from './AllCycles';
+
+const PlusButton = styled(TouchableOpacity)`
```
ctperry0301: Maybe PlusButton Could be a component?

components/cycles/AllCycles.jsx
```diff
@@ -0,0 +1,44 @@
...
+      onIconPress={item.onIconPress}
+      onPress={item.onPress}
+      color={item.color}
+    />
+  );
+
+  return (
+    <View>
+      <Title>Cycles</Title>
+      <FlatList
```
ctperry0301, line 30: Does FlatList allow you to scroll in the case a few more cycles are added?

App.js
```diff
@@ -9,8 +9,31 @@ import Calendar from './components/calendar/Calendar';
 import Home from './components/home/Home';
 import Profile from './components/profile/Profile';
 import Workouts from './components/workouts/Workouts';
+import Cycles from './components/cycles/Cycles';
 import { configureFonts, DefaultTheme, Provider as PaperProvider } from 'react-native-paper';
-
+import { Entypo } from '@expo/vector-icons';
+import { AppLoading } from 'expo';
+import {
+  useFonts,
```
ctperry0301, line 17: Not a Huge Deal, but we should get rid of fonts we wont use as each font slightly increase the time the app takes to load. (Like we probably arent ever going to use the italic ones or the super thin and thick ones)

ctperry0301 (dismissed):
Looks really good!

components/cycles/Cycles.jsx
jpoist97: Yeah good idea, I'll go and add that

components/cycles/AllCycles.jsx
jpoist97: Yep, it'll scroll

App.js
jpoist97: Yeah good point, should we just leave them all for now and throw a task up for once we have design finalized to remove extras?

finlaylp: Yeah agreed, let's make that a new task and edit out the ones we won't use later on, it isn't very urgent.

components/cycles/Cycles.jsx
```diff
@@ -0,0 +1,53 @@
...
+   padding: 10px;
+`;
+
+const Buttontext = styled.Text`
+   font-family: 'Montserrat_500Medium';
+   font-size: 18px;
+`;
+
+export default () => {
+  const items = [
```
finlaylp, line 14: Is this temporary or are these gonna be default placeholder cycle cells?

finlaylp (dismissed):
Looks great.

components/cycles/Cycles.jsx
jpoist97: These are temporary, they'll be replaced by a call to the database. I don't think we should have placehodler cells if a user does not have any cycles

finlaylp: Great was thinking the same thing

ctperry0301 (approved):

PR #11, opened 2020-10-30
[DEV-12] Implement Workout View Page

jpoist97:
Implemented Workout View page and also fixed some minor style bugs in Cycles View page

IMG_3729
IMG_3730

### DIFF
--- a/components/cycles/Cycles.jsx
+++ b/components/cycles/Cycles.jsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { SafeAreaView, } from 'react-native';
+import { SafeAreaView } from 'react-native';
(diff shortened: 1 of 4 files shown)


ctperry0301 (approved):

rohithdara:
Looks great, my only style issue would be if there are an odd number of workouts, the workout button gets lost in the background at the bottom.

Also, we should think about how many characters we should limit the workout to be because anything more than what "Back & Biceps" looks like may be too much.

jpoist97:
Yeah @rohithdara I'm using the same color from the mockups for the button, I think the background color of our app isn't pure white which is why its harder to see. And yeah I think we should limit workout names to ~15 chars

PR #13, opened 2020-10-31
Dev 11 firebase v3

finlaylp:
Not sure why it keeps adding to the package lock, but that happens no matter how I install firebase whether that's from expo or npm.


finlaylp:
I was thinking that if the package lock changes are a problem, I can just manually copy and paste the package lock from main and see if it still works.

jpoist97 (approved):

ctperry0301 (approved):

PR #14, opened 2020-11-04
[DEV-6] Create New Workout Page UI

ctperry0301:
Screen Shot 2020-11-03 at 4 36 08 PM

### DIFF
--- a/components/home/WorkoutContainer.jsx
+++ b/components/home/WorkoutContainer.jsx
@@ -1,9 +1,19 @@
+/* eslint-disable */
 import * as React from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Button, Subheading } from 'react-native-paper';
 
 const buttonStartLocation = 28;
 const textStartLocation = 30;
+
+const styles = StyleSheet.create({
+  startButton: {
+    backgroundColor: '#5DB075',
+    borderRadius: 20,
+    position: 'absolute',
+    right: 15,
+  },
+});
 
 export default (props) => (
   <View>
@@ -25,12 +35,3 @@ export default (props) => (
   </View>
 
 );
-
-const styles = StyleSheet.create({
-  startButton: {
-    backgroundColor: '#5DB075',
-    borderRadius: 20,
-    position: 'absolute',
-    right: 15,
-  },
-});
--- a/components/home/WorkoutContainer.jsx
+++ b/components/home/WorkoutContainer.jsx
@@ -1,19 +1,9 @@
-/* eslint-disable */
 import * as React from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Button, Subheading } from 'react-native-paper';
 
 const buttonStartLocation = 28;
 const textStartLocation = 30;
-
-const styles = StyleSheet.create({
-  startButton: {
-    backgroundColor: '#5DB075',
-    borderRadius: 20,
-    position: 'absolute',
-    right: 15,
-  },
-});
 
 export default (props) => (
   <View>
@@ -35,3 +25,12 @@ export default (props) => (
   </View>
 
 );
+
+const styles = StyleSheet.create({
+  startButton: {
+    backgroundColor: '#5DB075',
+    borderRadius: 20,
+    position: 'absolute',
+    right: 15,
+  },
+});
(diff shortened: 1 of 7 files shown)


components/workouts/CreateWorkout.jsx
```diff
@@ -0,0 +1,48 @@
...
+`;
+
+export default () => {
+  const [name, setName] = React.useState('Set Workout Name');
+
+  return (
+    <View style={{ height: '100%' }}>
+      <View>
+        <TitleTextInput
+          onChangeText={() => setName(name)}
```
finlaylp: Shouldn't name be a parameter here

finlaylp (dismissed):
Solid

rohithdara (dismissed):
looks good to me

components/utils/FinishButton.jsx
```diff
@@ -0,0 +1,43 @@
...
+   padding-left: 15px;
+   padding-right: 15px;
+`;
+
+const FinishButton = (props) => {
+  const {
+    onPress, style,
+  } = props;
+
+  const Buttontext = styled.Text`
```
jpoist97: nit: Move this outside of the FInishButton function since it doesn't use props

components/workouts/CreateWorkout.jsx
```diff
@@ -0,0 +1,48 @@
...
+          value={name}
+          maxLength={40}
```
jpoist97: Should we make this ~15 chars, I think that's about how many will fit on two lines of the workout cards

components/workouts/CreateWorkout.jsx
```diff
@@ -0,0 +1,48 @@
...
+        />
+        {/* Maybe make maxLength less, so workout name can't get cut off
+      prevents them from having long names of workouts */}
+
+      </View>
+      <AddFinishButton onPress={() => alert('Workout Created')} />
+      {/* This is where a list of set workout details components will go */}
```
jpoist97, line 54: This won't be set workout details components anymore right? This will be those "set x reps" components which I think should be a part of this task

components/workouts/Workouts.jsx
```diff
@@ -46,10 +49,21 @@ export default () => {
     items.push({ color: '#00000000', name: '', displayEllipses: false });
   }
 
-  return (
+  const Stack = createStackNavigator();
+  return ( // This allows you to access 2 different pages on same navigation tab. (i.e. Workouts)
```
jpoist97, line 53: @shriyan44 is this the functionality we wanted here?, also if it is I think we should include a back button too.

jpoist97 (changes requested):
Current code looks good, I left a question for Shriya about the design of this page, and then the only thing I think is missing from here is that "sep x reps" list of components which would go where you commented about set details component

components/workouts/CreateWorkout.jsx
ctperry0301: yup

ctperry0301:
ezgif com-gif-maker

finlaylp (dismissed):

components/home/WorkoutContainer.jsx
```diff
@@ -1,10 +1,20 @@
+/* eslint-disable */
```
jpoist97: I know this is giving linting issues but rohith is changing this component so this might create unnecessary merge conflicts

components/workouts/CreateWorkout.jsx
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable react/prop-types, no-shadow */
```
jpoist97: I think this is an important eslint check, I don't think we should disable this. i'll make a comment on how you can fix it

components/workouts/CreateWorkout.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <View style={{ height: '100%' }}>
+      <View>
+        <BackButton onPress={() => navigation.navigate('Workouts')}>
+          <AntDesign name="left" size={30} color="black" />
+        </BackButton>
+        <TitleTextInput
+          style={{ borderColor: name ? 'transparent' : 'black' }}
+          onChangeText={(name) => setName(name)}
```
jpoist97: To fix the linting rename this to something else. e.g. "newName"

jpoist97 (commented):
Looks good, just the small comment about the no-shadow thing. Also I prefer the textbox to keep the underline but that's just my preference (@shriyan44)

jpoist97 (approved):

PR #15, opened 2020-11-04
Dev 1 signup

finlaylp:
If there is a user logged in on app bootup, home screen will be presented, if not a signup screen will be prompted. I haven't added navigation from signup yet tho, so after successful signup you'll have to quit the app to be put to the home screen again (for now).


finlaylp:
Broken beyond repair. Some real nasty error is preventing simulator from working, I copied files and am gonna start a new branch

PR #17, opened 2020-11-06
[DEV-9] Log workout component no animations

jpoist97:
Implemented log workout component with no animations outside of showing the trash icon when swiping left on a set.


jpoist97:
To test add this code to Calendar component

`import LogWorkout from '../workouts/LogWorkout';`

```
<LogWorkout
    name="Pull A"
    exercises={[{
      name: 'Bicep Curls',
      sets: [{
        weight: '30', reps: '8',
      }, {
        weight: '30', reps: '8',
      }, {
        weight: '30', reps: '8',
      }, {
        weight: '30', reps: '8',
      }],
    }, {
      name: 'Rows',
      color: '#6D8DFF',
      sets: [{
        weight: '150', reps: '12',
      }, {
        weight: '150', reps: '12',
      }, {
        weight: '150', reps: '12',
      }, {
        weight: '150', reps: '12',
      }],
    }, {
      name: 'Reverse Fly',
      color: '#9D8DFF',
      sets: [{
        weight: '25', reps: '20',
      }, {
        weight: '25', reps: '20',
      }, {
        weight: '25', reps: '20',
      }, {
        weight: '25', reps: '20',
      }],
    }]}
  />

```

jpoist97:
IMG_3747

components/utils/PlusButton.jsx
```diff
@@ -13,11 +13,11 @@ const ButtonContainer = styled(TouchableOpacity)`
 
 const PlusButton = (props) => {
   const {
-    onPress, title, size, style,
+    onPress, title, size, style, font,
   } = props;
 
   const Buttontext = styled.Text`
-      font-family: 'Montserrat_500Medium';
+      font-family: ${font};
```
ctperry0301, line 20: Im a little unsure as to why font is a prop? Is the font something that might change depending on where it occurs? Is this just a good practice?

components/workouts/ExerciseDetails.jsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <Container style={{ backgroundColor: color }}>
+      <ExerciseName>{name}</ExerciseName>
+      <LogHeader />
+      {items.map((set, index) => (
+        <SetDetails
+          prevWeight={set.prevWeight}
+          prevReps={set.prevReps}
+          weight={set.weight}
+          reps={set.reps}
```
ctperry0301: Did we decide on a way to log time for timed workouts? We might want that in SetDetails

components/workouts/ExerciseDetails.jsx
```diff
@@ -0,0 +1,78 @@
...
+          onWeightChange={updateWeight(index)}
+          onCompletedPress={updateCompleted(index)}
+          onSetDelete={onSetDelete(index)}
+          completed={set.completed}
+          setNumber={index + 1}
+          key={index}
+        />
+      ))}
+      <ButtonFlexContainer>
+        <AddSetButton size={16} title="Add Set" onPress={onSetAdd} />
```
ctperry0301: I think the default size for the plus and finish button is 18. We should keep these consistent

components/workouts/LogWorkout.jsx
```diff
@@ -0,0 +1,132 @@
...
+      />
+    </SafeAreaView>
+  );
+};
+
+LogWorkout.propTypes = {
+  exercises: PropTypes.array.isRequired,
+  name: PropTypes.string.isRequired,
+};
+
```
ctperry0301: The app breaks if you delete all the sets, then try to add a set.

I dont think we should let the user delete a set if there is only 1 set left. 

Also we should start to think about if we want the user to be able to customize their workout during a workout, (i.e. The row machine is occupied, so they want to switch it out with pull ups)

ctperry0301 (commented):
Your a coding god lol. You should demo/walk us through your code the next time we meet. Theres a lot going on and everything looks really good. I dont really understand everything that's going on and I think it would probably be helpful for the rest of us to walk us through it.

components/workouts/LogWorkout.jsx
jpoist97: Ah good catch, yeah I agree we shouldn't let them delete it if there is one set left

components/utils/PlusButton.jsx
jpoist97: I added it bc on the mockups the font for the add set button is roboto not montseratt. I didn't use it bc I thought Montseratt looks better, but I left it in case we want to change it in the future. Can probably remove it it's not hard to add back in later if we need it

components/workouts/ExerciseDetails.jsx
jpoist97: Oh yeah I forgot about that, I'll update this pr with that

components/workouts/LogWorkout.jsx
jpoist97: Also yeah we should allow them to switch out, I think we need to add something in the UI for this functionality, maybe the ellipses on the top right of each ExerciseDetails component? @shriyan44

components/workouts/ExerciseDetails.jsx
jpoist97: This button is 16 bc the rest of the text in the SetDetails Component is 16px, We could bump it to 18px but I think it'd look a little weird @shriyan44 what u think

components/workouts/LogWorkout.jsx
jpoist97: IMG_3753

finlaylp (approved):
This is dope I'm impressed. For future PRs if they get this big it might make review a lot easier to just leave a few comments here and there

PR #19, opened 2020-11-09
Dev 11 auth

finlaylp:
opens home if already logged in, login screen if not; most auth functionality done (logging in, swapping from signup to login); no email confirmation on signup yet; no forgot password;


App.js
```diff
@@ -107,40 +112,20 @@ export default function App() {
     console.log("Firebase setup already complete.")
   }
 
+  const user = firebase.auth().currentUser;
+  var startupScreen = 'Login'
```
rohithdara, line 143: just curious why this is a var instead of const

App.js
```diff
@@ -107,40 +112,20 @@ export default function App() {
...
-          <Tab.Screen name="Calendar" component={Calendar} />
-          <Tab.Screen name="Profile" component={Profile} />
-        </Tab.Navigator>
+        <Stack.Navigator initialRouteName={startupScreen}>
+          <Stack.Screen name="Login" component={login} options={{headerShown:false}}/>
+          <Stack.Screen name="Signup" component={signup} options={{headerShown:false}}/>
+          <Stack.Screen name="Root" component={Root} options={{headerShown:false}} />
+        </Stack.Navigator>
       </NavigationContainer>
       </PaperProvider>
```
rohithdara, line 157: tab inconsistency I think (like </PaperProvider> should be 1 tab back)

components/authentication/login.jsx
```diff
@@ -1,163 +1,144 @@
...
-export default class login extends React.Component {
-    state={
-        email:"",
-        password:"",
-        showPassword:true,
-    }
+export default ({ navigation }) => {
+  const [email, setEmail] = React.useState('');
+  const [password, setPassword] = React.useState('');
+  const [showPassword, setShowPassword] = React.useState(true);
```
rohithdara: Should the default for showpassword be true?

rohithdara (commented):
Looks good to me my guy. Would be helpful if you included screenshots of what the screens themselves look like

App.js
finlaylp: because it isn't constant, it gets reassigned a couple lines down

components/authentication/login.jsx
finlaylp: I guess I should change it, but true means that the password is hidden.

rohithdara: oh then true makes sense, I assumed the opposite

rohithdara: but yeah I guess variable name wise that is confusing

finlaylp: yeah I'll reverse

jpoist97 (commented):
Looks good, I'm trusting you've already tested all of the login/signup stuff cause I can't figure out how to log out and do it again hahaha. Good shit tho. Only thing is that when I'm typing in my password on my iphone I can't actually see the textbox im typing in since the keyboard covers it. I used KeyboardAwareFlatList for this in the LogWorkout component and im sure there's something similar that you can use for this

components/authentication/signup.jsx
```diff
@@ -1,169 +1,154 @@
...
-                            >
-                    </Image>
-                    <TextInput style={styles.nameInput}
-                                selectionColor={"#5DB075"}
-                                placeholder={'Name'}
-                                mode={'flat'}
-                                underlineColor={'transparent'}
-                                >
+          default:
+            console.log(errorMessage);
```
ctperry0301, line 37: Add a way to display the error message to the user.

components/authentication/signup.jsx
```diff
@@ -1,169 +1,154 @@
...
+    <SafeAreaView style={{ flex: 1 }}>
+      <View style={styles.container}>
+        <Image
+          style={styles.icon}
+          source={require('../../assets/icon.png')}
+        />
+        <TextInput
+          style={styles.input}
+          selectionColor="#A192FF"
+          placeholder="Name"
```
ctperry0301: Maybe change placeholder to first name, cause that's all we care about

components/authentication/login.jsx
```diff
@@ -1,163 +1,144 @@
...
+        }
+      });
+  }
 
-    signUpPressed = () => {
-        alert("sign up pressed")
-    }
+  function signupPress() {
+    navigation.navigate('Signup');
+    console.log("Navigating to signup...")
```
ctperry0301, line 33: Display message to user if they have an incorrect password.

ctperry0301 (dismissed):
Looks really good. After fixing the comments, should be good to go ahead and merge.

components/authentication/login.jsx
finlaylp: This part is just if they don't have an account yet and navigate to the signup page from the login page, so no info is saved

ctperry0301 (approved):
Looks good to me

PR #20, opened 2020-11-10
Implement workout section list

jpoist97:
Implemented section list for workout component and EllipsisPopup

We also need to decide how we want to do the colors now that this list is sorted and split into sections


jpoist97:
IMG_3767

jpoist97:
Screenshot of popup
IMG_3768

components/workouts/AllWorkouts.jsx
```diff
@@ -11,29 +13,100 @@ const Title = styled.Text`
...
+    {right ? (
+      <WorkoutCard
+        name={right.name}
+        subtext={right.subtext}
+        displayEllipses={right.displayEllipses}
+        deleteWorkout={right.deleteWorkout}
+        onPress={right.onPress}
+        color={right.color}
+        key={right.name + right.subtext}
+      />
```
ctperry0301, line 82: I don't have a problem with the "left cards & right cards" way. It's pretty easy to understand and it works as expected. 
That being said, why did you do it like this as opposed to just using styling and numColumns?

ctperry0301 (dismissed):
Looks good.

components/workouts/AllWorkouts.jsx
jpoist97: Unless I missed something in the documentation, there isn't numColumns for the alphabet list component I used for this. It also wasn't there for any of the other options I looked at, although documentation wasn't always great.

ctperry0301 (approved):

components/workouts/AllWorkouts.jsx
```diff
@@ -11,29 +13,100 @@ const Title = styled.Text`
...
+    } else {
+      accumulator[bucket].push(item);
+    }
+
+    return accumulator;
+  }, {});
+
+  // Pair up each bucket of data
+  const pairedBucketData = _.mapValues(bucketData, (data) => {
+    const pairData = [];
```
shriyan44, line 50: What does pairing up mean here?

shriyan44 (approved):
good job troni

PR #22, opened 2020-11-12
Dev 7 Add Exercise UI

rohithdara:
- got search filtering working
- got exercise list with clickable button working
- only bug seems to be sometimes the button on the exercise doesn't change when clicked after search filtering occurred.. weird because it works sometimes and not others
- added exercises saves to a list but doesn't connect to create a workout yet


rohithdara:
Screen Shot 2020-11-12 at 3 14 26 PM

components/workouts/CreateWorkout.jsx
```diff
@@ -52,7 +52,7 @@ export default ({ navigation }) => {
       <AddFinishButton onPress={() => alert('Workout Created')} />
       {/* Finish Button will take u back to workouts and add workout to list */}
       {/* This is where a list of set workout details components will go */}
-      <AddCycleButton title="Exercise" size={18} onPress={() => alert('Add Exercies')} />
+      <AddCycleButton title="Exercise" size={18} onPress={() => navigation.navigate('Add Exercises')} />
```
jpoist97, line 55: We want this to be a "popover" right? Which means we need to switch this to a full screen modal. I'm working on an example of this for the LogWorkouts component

components/workouts/ExerciseItem.jsx
```diff
@@ -0,0 +1,80 @@
...
+          />
+        )
+        : (
+          <AddButton
+            name="pluscircleo"
+            size={22}
+            color="#CAB0FF"
+            onPress={onPress}
+          />
+        )}
```
jpoist97: nit: Only thing that changes between these two buttons is the name prop, so you can just use the same tag an conditionally pass name

components/workouts/ExerciseItem.jsx
```diff
@@ -0,0 +1,80 @@
...
+    margin: 0px 25px 0px 0px;
+`;
+
+const ExerciseItem = (props) => {
+  const {
+    name, subtext, selected, onPress,
+  } = props;
+
+  return (
+    <View style={{
```
jpoist97, line 36: nit: use styled-components for this

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+  const { items } = props;
+
+  const parsedItems = parseItems(items);
+  const [search, setSearch] = React.useState('');
+  const [filteredDataSource, setFilteredDataSource] = React.useState([]);
+  const [masterDataSource, setMasterDataSource] = React.useState([]);
+  React.useEffect(() => {
+    setFilteredDataSource(parsedItems);
+    setMasterDataSource(parsedItems);
+  }, []);
```
jpoist97: If the only purpose of this useEffect is to set the initial state you can do this, problem with the way you're doing it is that you're passing it an array initially which is not what the data prop in alphabetSectionList expects so we get a warning and sometimes it'll act weird
```suggestion
  const [filteredDataSource, setFilteredDataSource] = React.useState(parsedItems);
  const [masterDataSource, setMasterDataSource] = React.useState(parsedItems);
```

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+  const searchFilterFunction = (text) => {
+    // Check if searched text is not blank
+    if (text) {
+      // Inserted text is not blank
+      // Filter the masterDataSource
+      // Update FilteredDataSource
+      const firstLetter = text[0];
```
jpoist97: Just to be safe here should we make this upper case?

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+      if (masterDataSource[firstLetter]) {
+        const newData = masterDataSource[firstLetter].filter((item) => {
+          const itemData = item.name
+            ? item.name.toUpperCase()
+            : ''.toUpperCase();
```
jpoist97: Why make this uppercase?

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+          const textData = text.toUpperCase();
+          return itemData.indexOf(textData) > -1;
+        });
+        const finished = {};
```
jpoist97: let's rename this like filteredData, finished was a bad name I just picked randomly haha

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+        finished[firstLetter] = newData;
+        setFilteredDataSource(finished);
+        setSearch(text);
+      } else {
+        setFilteredDataSource(masterDataSource);
```
jpoist97: This is the case where there is no match right? Should we set this to an empty object instead?

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+      }
+    } else {
+      // Inserted text is blank
+      // Update FilteredDataSource with masterDataSource
+      setFilteredDataSource(masterDataSource);
+      setSearch(text);
+    }
+  };
+  const [exerciseCount, setExerciseCount] = React.useState(0);
+  const [addedExercises] = React.useState([]);
```
jpoist97: Let's move this to the top of your functional component with the other calls to useState

components/workouts/ExerciseList.jsx
```diff
@@ -0,0 +1,154 @@
...
+      selected={item.selected}
+      onPress={() => {
+        const temp = { ...masterDataSource };
+        const { selected } = temp[item.name[0]][index];
+        temp[item.name[0]][index].selected = !selected;
+        setMasterDataSource(temp);
+        (temp[item.name[0]][index].selected === true) ? setExerciseCount(exerciseCount + 1)
+          : setExerciseCount(exerciseCount - 1);
+        (temp[item.name[0]][index].selected === true) ? addedExercises.push(item)
+          : addedExercises.splice(addedExercises.indexOf(item), 1);
```
jpoist97, line 118: Is there a reason you chose to constantly update this rather than calculate it at the end?

jpoist97 (commented):
Looks good Rohith! I made a couple comments, but looking good overall. I also couldn't reproduce the bug you mentioned about not updating the button

Oh and one other thing is when I ran npm i on my computer with your branch it gave me an error in the package-lock, before you push anything else try deleting package-lock and running npm i again

components/workouts/ExerciseItem.jsx
rohithdara: good idea, changed

components/workouts/ExerciseList.jsx
rohithdara: I just thought it would be easy to do it this way without using too much rendering power (not sure though). If there is a need later as we are tidying up code for it to be a function, then I can do that.

shriyan44:
Didnt really look at ur code since Justin did, but just one nitpicky thing can u make the search bar line up with the exercise title and add button and move those three items a bit up as mentioned before

components/workouts/ExerciseList.jsx
rohithdara: Yep good call, fixed

components/workouts/ExerciseList.jsx
jpoist97: Yeah this is fine for now, we can go back and clean it up after our presentation

jpoist97 (commented):
Looks good, I think this part is ready to merge with Cole's code and get it all hooked up

PR #23, opened 2020-11-13
Dev 8 set workout details page

ctperry0301:
ezgif com-gif-maker


components/setWorkoutDetails/SetWorkoutDetailsCard.jsx
```diff
@@ -0,0 +1,155 @@
...
+`;
+
+const SetWorkoutDetailsCard = (props) => {
+  const {
+    color, name, displayEllipses,
+  } = props;
+  const [sets, setSets] = React.useState('');
+  const [reps, setReps] = React.useState('');
+  const [seconds, setSeconds] = React.useState('');
+  const [isReps, setIsReps] = React.useState(true);
```
jpoist97: I think you need to move your state a little higher and into the CreateWorkout component so you can access this information when the finish button is clicked and send it to the database

jpoist97 (changes requested):
Looks good, only thing I see is you need to move the state of the SetWorkoutDetailsCard into your itemState in CreateWorkout. Kind of following the pattern I used for LogWorkout. Lmk if you have questions about that and I can walk you through it. Otherwise looks good!

PR #28, opened 2020-11-17
Edit cycle order

finlaylp:
Vertically Centered:
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-17 at 10 45 18
The all caps workout names will change, I just used the ID for it for now. The other stuff is lined up with the data structure so far.


components/cycles/CycleOrder.jsx
```diff
@@ -0,0 +1,128 @@
...
+          <SubTitle>Exercise</SubTitle>
+          <SubTitle>Sets x Reps</SubTitle>
+          <SubTitle>Previous</SubTitle>
+        </RowHeader>
+        <FlatList
+          data={workoutItems}
+          keyExtractor={(item) => item.id}
+          renderItem={({ item }) => (
+            <RowContent>
+              <WorkoutViewText style={{ marginLeft: 10 }}>{item.id}</WorkoutViewText>
```
jpoist97, line 81: Can you add a todo comment saying this is where we switch it to exercise name once redux goes in

components/cycles/CycleOrder.jsx
```diff
@@ -0,0 +1,128 @@
...
+};
+
+CycleOrderCard.defaultProps = {
+  name: 'Cycle Name',
+  muscleGroups: '',
+  color: '#CAB0FF',
+  exercises: [],
+};
+
+export default CycleOrderCard;
```
jpoist97: Can we rename this file to something more descriptive, maybe like "ExpandableCycleCard"

components/cycles/CycleOrder.jsx
```diff
@@ -0,0 +1,128 @@
...
+        background-color: ${color};
+        borderRadius: 10px;
+        box-shadow: 1px 5px 2px gray;
+        marginBottom: 10px;
+        marginTop: 10px;
+        padding: 10px;
+    `;
+
+  const ExercisesView = (items) => {
+    const workoutItems = items.items;
```
jpoist97: The function that represents a component should always take in "props", also can we rename items and workoutItems to something more descriptive, say exercises.
```suggestion
  const ExercisesView = (props) => {
    const { exercises } = props;
```

components/cycles/CreateCycle.jsx
```diff
@@ -3,9 +3,75 @@ import { AntDesign } from '@expo/vector-icons';
 import * as React from 'react';
 import { View } from 'react-native';
 import styled from 'styled-components/native';
+import FullCycleOrder from './FullCycleOrder';
 import FinishButton from '../utils/FinishButton';
 import PlusButton from '../utils/PlusButton';
 
+const workouts = [
```
jpoist97: **This needs to be state**

jpoist97 (changes requested):
Looks good for the most part, one thing you're missing is you need to add state to CreateCycle, we need this component to rerender whenever workouts changes. Should be a quick change since the state shouldn't be too complicated, basically just need to make workouts stateful and should be good

One small styling thing I noticed, can we vertically center the text in the expanded cycle card, that way like romanian deadlift will look a bit more uniform

jpoist97:
Looks like this should be ready to go in soon, @shriyan44 how close is your part? I think once both parts are ready we should make a new branch like we did with Cole and Rohith's code and connect your two components before merging with main.

components/cycles/CreateCycle.jsx
finlaylp: I thought this was cole's part since I didn't make this file but I don't mind doing it

jpoist97: yeah can you do this part please

finlaylp: yep

jpoist97:
Looks good to merge w shriya's code once that is done

PR #30, opened 2020-11-17
Fix add exercise ui

rohithdara:
Refactored Exercises to a new folder
Changed icons from + to check mark
Changed the clickable area for adding an exercise to be larger


rohithdara:
I think some other files got changed when I ran autofix with lint, but all was minor syntax changes

jpoist97 (approved):

PR #31, opened 2020-11-17
Add support for timed exercises in logworkouts

jpoist97:
IMG_3793

### DIFF
--- a/components/logWorkout/LogHeader.jsx
+++ b/components/logWorkout/LogHeader.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import styled from 'styled-components/native';
+import PropTypes from 'prop-types';
 
 const Container = styled.View`
   flex-direction: row;
@@ -17,16 +18,26 @@ const StyledText = styled.Text`
   text-align: center;
 `;
 
-export default () => (
+const LogHeader = ({ type }) => (
   <Container>
     <StyledText style={{ width: 30 }}>Set</StyledText>
 
     <StyledText style={{ width: 70 }}>Previous</StyledText>
 
-    <StyledText style={{ width: 70 }}>lbs</StyledText>
+    {type === 'WEIGHT' ? <StyledText style={{ width: 70 }}>lbs</StyledText> : <StyledText style={{ width: 70 }}>Sec</StyledText> }
 
     <StyledText style={{ width: 70 }}>Reps</StyledText>
 
     <StyledText style={{ width: 40 }}>Done</StyledText>
   </Container>
 );
+
+LogHeader.propTypes = {
+  type: PropTypes.string,
+};
+
+LogHeader.defaultProps = {
+  type: 'WEIGHT',
+};
+
+export default LogHeader;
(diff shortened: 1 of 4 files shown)


rohithdara:
I'm confused as to the structure of this because I think there is a discrepancy somewhere. I thought it would make more sense to replace the reps with the time instead of the weight in case they wanted to have weighted timed exercises. If not, I think the 'previous' has the wrong unit because based on the current structure I think it should say 60 secs not 60 lbs. Let me know if I'm confusing myself or if what I'm saying makes sense

components/logWorkout/SetDetails.jsx
```diff
@@ -118,10 +118,10 @@ const SetDetails = (props) => {
       </SetNumber>
 
       <PrevWeight>
-        <StyledText>{prevWeight !== 'n/a' ? `${prevWeight} lbs.` : prevWeight}</StyledText>
+        <StyledText>{prevPerRep !== 'n/a' ? `${prevPerRep} lbs.` : prevPerRep}</StyledText>
```
rohithdara, line 121: I'm confused as to the structure of this because I think there is a discrepancy somewhere. I thought it would make more sense to replace the reps with the time instead of the weight in case they wanted to have weighted timed exercises. If not, I think the 'previous' has the wrong unit because based on the current structure I think it should say 60 secs not 60 lbs. Let me know if I'm confusing myself or if what I'm saying makes sense

components/logWorkout/LogHeader.jsx
```diff
@@ -17,16 +18,26 @@ const StyledText = styled.Text`
...
 
-export default () => (
+const LogHeader = ({ type }) => (
   <Container>
     <StyledText style={{ width: 30 }}>Set</StyledText>
 
     <StyledText style={{ width: 70 }}>Previous</StyledText>
 
-    <StyledText style={{ width: 70 }}>lbs</StyledText>
+    {type === 'WEIGHT' ? <StyledText style={{ width: 70 }}>lbs</StyledText> : <StyledText style={{ width: 70 }}>Sec</StyledText> }
```
rohithdara, line 27: Also small thing, but every header has first letter capitalized except for lbs. Maybe change to Lbs? Idk if that looks ugly lol

components/logWorkout/SetDetails.jsx
jpoist97: wait yeah you're right gdi hahaha

jpoist97: good catch

rohithdara (commented):
The code itself looks good other than the question I posed in one of the comments. Let's also make a note to look into making the seconds into a timer or something like that in the future.

components/logWorkout/LogHeader.jsx
jpoist97: Yeah Lbs kinda ugly, I thought it looked weird with lower case sec as well

jpoist97:
I'm gonna start over on a new branch cause it's harder to change it back and then change the sets thing up, but thanks for the review rohith

PR #33, opened 2020-11-17
added logout functionality

ctperry0301:
logout button in profile page logs user out and takes them back to login page


finlaylp (approved):
Nice

PR #34, opened 2020-11-18
Merge create cycle

jpoist97:
PR includes implementation for CreateCycles and AddWorkouts as well as the code to merge them together
IMG_3820


jpoist97:
Forgot to mention in the add workout screen the subtext won't show up, this will be fixed when we add redux

components/cycles/AddWorkouts/AddWorkouts.jsx
```diff
@@ -0,0 +1,230 @@
...
+            (temp[right.name[0].toUpperCase()][index].right.selected === true) ? setWorkoutCount(workoutCount + 1)
+              : setWorkoutCount(workoutCount - 1);
+            (temp[right.name[0].toUpperCase()][index].right.selected === true) ? addedWorkouts.push(right)
+              : addedWorkouts.splice(addedWorkouts.indexOf(right), 1);
+          }}
+          displayAddButton={right.displayAddButton}
+          color={right.color}
+          key={right.name + right.subtext}
+        />
+      ) : (
```
finlaylp, line 196: Might want to make a more empty one instead of using the same component, bc this component has a box shadow which I think would still show up.

finlaylp (approved):
Looks good, not sure if the box shadow does show up on an empty SelectableWorkoutCard but if it does should prob change that

components/cycles/AddWorkouts/AddWorkouts.jsx
jpoist97: I just checked and it doesn't show the box shadow, you're right tho I feel like it should but it doesn't so we're good here hahah

components/cycles/CreateCycle.jsx
```diff
@@ -32,28 +38,53 @@ const AddCycleButton = styled(PlusButton)`
...
+        placeholder="Cycle Name"
+      />
+
       <AddFinishButton onPress={() => alert('Cycle Created')} />
-      {/* Finish Button will take u back to Cycles and add cycle to list */}
-      {/* This is where a list of workouts in the cycle will go. Should be
-      able to drag to change order */}
-      <AddCycleButton title="Workout" size={18} onPress={() => alert('Add Workout to Cycle')} />
-    </View>
+
```
ctperry0301, line 63: Should we add finish button functionality before the presentation?

jpoist97: Yeah I'll add that w redux, redux will also let us display the workout we create in allCycles view

components/cycles/CycleOrder.jsx
```diff
@@ -0,0 +1,131 @@
...
+    if (icon == 'chevron-down') {
+      setIcon('chevron-up');
+    } else {
+      setIcon('chevron-down');
+    }
+    setshowDetailed(!showDetailed);
+  }
+
+  return (
+    <FullBody>
```
ctperry0301, line 102: I feel like it makes more since to expand the card if the card is pressed at all, not just if they hit the icon.

finlaylp: I asked you guys that when I was working on that branch and I thought we agreed on just the icon. I can go and change it if we wanna do that tho lol

components/cycles/CycleOrder.jsx
```diff
@@ -0,0 +1,131 @@
...
+      <CycleTitle>{name}</CycleTitle>
+      <IconButton
+        icon={icon}
+        color="white"
+        size={25}
+        onPress={showDetail}
+        style={{ position: 'absolute', alignSelf: 'flex-end', marginTop: 0 }}
+      />
+      {!showDetailed && <SubTitle>{muscleGroups}</SubTitle>}
+      {showDetailed && <ExercisesView exercises={exercises} />}
```
ctperry0301, line 112: IDK how I feel about the all CAPS for exercises. Also it might look better if the text is left-aligned

ctperry0301 (approved):

components/cycles/CycleOrder.jsx
jpoist97: The all caps will be fixed when redux goes in, that's just the exercise id for now

components/cycles/CycleOrder.jsx
ctperry0301: lets see what the group thinks when we meet

components/cycles/CycleOrder.jsx
finlaylp: ^ just temporary. And idk about left aligned, in figma they're center aligned so I just followed that. I can try it and see how it looks tho

PR #35, opened 2020-11-19
Cycle order draggable

finlaylp:
changed the cycle's workouts from a flatlist to a draggable flatlist, states update when a component is dragged.

### DIFF
--- a/components/cycles/FullCycleOrder.jsx
+++ b/components/cycles/FullCycleOrder.jsx
@@ -5,14 +5,22 @@ import ExpandableCycleCard from './CycleOrder';
 import DraggableFlatList from 'react-native-draggable-flatlist';
 
 const FullCycleOrder = (props) => {
-  const { workouts } = props;
+
+  const { 
+    workouts, passWorkoutList 
+  } = props;
 
   const [workoutList, setWorkoutList] = React.useState(workouts)
+
+  React.useEffect(() => {
+    passWorkoutList(workoutList)
+  }, [workoutList]);
+
   return (
     <DraggableFlatList
       data={workoutList}
       keyExtractor={(item, index) => item.id.toString() + index}
-      onDragEnd={({data}) => setWorkoutList(data)}
+      onDragEnd={({data}) => setWorkoutList(data) }
       renderItem={({ item, drag}) => (
         <ExpandableCycleCard
           drag={drag}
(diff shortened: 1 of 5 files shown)


ctperry0301 (commented):
Overall it works pretty well. I think it would nice to have an icon on the left side of the card to show that the card is draggable. Maybe one of these could look nice.

https://icons.expo.fyi/MaterialIcons/swap-vert
https://icons.expo.fyi/MaterialIcons/drag-handle

components/cycles/FullCycleOrder.jsx
```diff
@@ -2,16 +2,28 @@ import * as React from 'react';
...
+
+  const { 
+    workouts, passWorkoutList 
+  } = props;
+
+  const [workoutList, setWorkoutList] = React.useState(workouts)
+
+  React.useEffect(() => {
+    passWorkoutList(workoutList)
+  }, [workoutList]);
```
jpoist97, line 17: Can you explain what you're doing with passWorkoutList?

jpoist97 (commented):
Looks good overall, could you just explain what you're doing w the passWorkoutList stuff

components/cycles/FullCycleOrder.jsx
jpoist97: Ohhhh nvm I get it

jpoist97: Can we rename passWorkoutList to like "SetWorkoutList" or updateWorkoutList

finlaylp: yeah useEffect is a hook that basically triggers when state gets updated as a side effect to changing the state. So when the state in this file for the input changes, which is the [workoutList] part, it triggers that which in turn passes the new workoutList state to the CreateCycle file. Bad explanation but that's about all I understand about it

finlaylp: setworkoutlist is what I use to set the workoutlist state variable in the hook so that might make it confusing

jpoist97: Yeah I think the proper way to do this is with onDragEnd but this looks like it works for now, I'm fine w this going in as is

jpoist97 (approved):

components/cycles/FullCycleOrder.jsx
finlaylp: That's what I originally tried, but it threw a warning about mounting a component or something and when I looked into it someone online recommended using the useEffect hook since that is what it's for

PR #38, opened 2020-12-15
login and signup tuned up; other files are just from lint autofix

finlaylp:
changed some things up with login and signup, they don't have any more lint errors. lint autofix also changes some other files but I personally didn't touch them


jpoist97 (approved):
Looks good to me!

PR #40, opened 2021-01-04
Capitilize Folders (bug fix from refactor)

ctperry0301:
When the folders names were changed in JP's refactor, these changes weren't pushed to git (For some reason, git doesn't track folder name changes if the only change is capitalization). This caused import issues in our app. This fixes that.


finlaylp (approved):
Thanks for this. Login and singup still not capitalized, I assume that doesn't cause any issues?

ctperry0301:
> Thanks for this. Login and singup still not capitalized, I assume that doesn't cause any issues?

The issue was with the folders, not the jsx files themselves, so the fix was updating the "authentication" to "Authentication"

PR #41, opened 2021-01-05
Make add exercises draggable

ctperry0301:


finlaylp (approved):
Looks good and works well, nice

jpoist97 (commented):
Code looks good and it works great. I keep getting a warning when I run it

"Animated: 'useNativeDriver' was not specified. This is a required option and must be explicitly set to 'true' or 'false'"

Not really sure what this warning means, did you run into this when u were doing this task?

ctperry0301:
> Code looks good and it works great. I keep getting a warning when I run it
> 
> "Animated: 'useNativeDriver' was not specified. This is a required option and must be explicitly set to 'true' or 'false'"
> 
> Not really sure what this warning means, did you run into this when u were doing this task?

Yeah, that warning seems to be coming from the deprecated package that i used. It can be found here [https://openbase.io/js/react-native-keyboard-aware-view/contributors](url)

However, I did a decent amount of looking for a replacement package and the only other one i could find [https://www.npmjs.com/package/react-native-keyboard-aware-scroll-view](url) did not give me the correct functionality.

PR #43, opened 2021-01-08
Removed unused fonts

jpoist97:
Removed all unused fonts, will be able to add them back in if we need them later


finlaylp (approved):

shriyan44 (approved):
good work boss

PR #44, opened 2021-01-10
Auth final

finlaylp:
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-10 at 15 21 44
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-10 at 15 21 55
Fixed up some random stuff in auth pages; added live updating for validating name, email, and password on signup; set user displayname on signup.


components/Authentication/signup.jsx
```diff
@@ -60,46 +59,90 @@ const ShowText = styled.Text`
...
+    setValidEmail(!ValidateEmail(email));
+  }, [email]);
+
+  React.useEffect(() => {
+    if (isFirstRunPassword.current) {
+      isFirstRunPassword.current = false;
+      return;
+    }
+    setValidPassword(!(password.length > 5));
+  }, [password]);
```
jpoist97, line 116: Could you explain what you're trying to do with this code, I'm a bit confused by it

jpoist97 (commented):
Also is the forgot password done? I don't see it when I run the app

components/Authentication/signup.jsx
```diff
@@ -60,46 +59,90 @@ const ShowText = styled.Text`
...
 
   function loginPress() {
     navigation.navigate('Login');
   }
 
-  firebase.auth().onAuthStateChanged(() => {
-    if (firebase.auth().currentUser) {
-      navigation.navigate('Root');
-    }
-  });
```
jpoist97, line 101: Also I think removing this makes it so we don't auto-login if we're already logged in

components/Authentication/signup.jsx
finlaylp: isFirstRun... tells if it is the first time running, since they're connected to useEffect it runs in the beginning when there is no text in the input. So it makes the error messages only show if it isn't the first time running

components/Authentication/signup.jsx
finlaylp: onAuthStateChanged only triggers on sign in or sign out, so it was redundant because when the login sign in resolves it will do what this does already. The auto-login happens through App.js, it won't even send them to signup or login if the device is logged in.

components/Authentication/signup.jsx
finlaylp: Oops forgot about forgot password. I'll put that in for login rq.

PR #46, opened 2021-01-11
Initialize App data from Firestore

jpoist97:
Redux is now initialized from Firestore. We need to do a bit more work on the DraggableWorkoutList since it was displaying the exerciseIds and not the names. Also I left color out of Firestore for now since we haven't decided what we want to do for that. But I'll add that in a later PR if that's what we decide to go with.

### DIFF
--- a/components/CreateCycle/ExpandableWorkoutCard.jsx
+++ b/components/CreateCycle/ExpandableWorkoutCard.jsx
@@ -76,12 +76,12 @@ const ExpandableWorkoutCard = (props) => {
         <FlatList
           data={exercises}
           // TODO: rename to exercise name
-          keyExtractor={(item) => item.id}
+          keyExtractor={(item) => item.exerciseId}
           renderItem={({ item }) => (
             <RowContent>
-              <WorkoutViewText style={{ marginLeft: 10 }}>{item.id}</WorkoutViewText>
+              <WorkoutViewText style={{ marginLeft: 10 }}>{item.exerciseId}</WorkoutViewText>
               {/* TODO: center the sets x reps around the x so double digit numbers dont look bad */}
-              <WorkoutViewText>{`${item.sets.length}x${item.sets[0].duration}`}</WorkoutViewText>
+              <WorkoutViewText>{`${item.sets.length}x${item.sets[0].reps || item.sets[0].time}`}</WorkoutViewText>
               <WorkoutViewText>{`${item.sets[0].weight} lbs.`}</WorkoutViewText>
             </RowContent>
           )}
--- a/components/CreateCycle/ExpandableWorkoutCard.jsx
+++ b/components/CreateCycle/ExpandableWorkoutCard.jsx
@@ -3,7 +3,9 @@ import { FlatList } from 'react-native-gesture-handler';
 import { IconButton } from 'react-native-paper';
 import styled from 'styled-components/native';
 import PropTypes from 'prop-types';
-import * as React from 'react';
+import React from 'react';
+import { useSelector } from 'react-redux';
+import _ from 'lodash';
 
 const CycleTitle = styled.Text`
    color: #FFFFFF;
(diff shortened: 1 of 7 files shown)


finlaylp (dismissed):
Looks good. So this works assuming the data has already been initialized for the user right? Which is the task I'm assigned

jpoist97:
> for the user right? Which is the task I'm assigned

Yeah, I guess we should either hold off on merging this PR in, or we create like a test account with some test data so when we're developing the app will be filled out still.

components/Home/Home.jsx
```diff
@@ -27,17 +29,87 @@ const welcomeName = 'Shriya';
 
 export default () => {
   useEffect(() => {
-    // This is where we would hit our database, but for now we'll have fake data
-    console.log('Home: Initializing Workout store');
-    dispatch({ type: INITIALIZE_WORKOUTS, workouts: DBWorkoutResponse });
+    const initializeDatabase = async () => {
+      // Get the current logged in user id
+      const currentUser = firebase.auth().currentUser.uid;
+      // const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';
```
jpoist97, line 105: Forgot to mention this, but when you're running this locally before Fin's task to update firebase when a user is created, comment out line 34 and uncomment line 35 and you'll be able to get the test data in my personal account

If you don't do this the app will error out bc there will be nothing in it

components/LogWorkout/LogWorkout.jsx
```diff
@@ -23,18 +23,22 @@ const TitleText = styled.Text`
...
+  const exerciseType = exercise.sets[0].reps ? 'REPS' : 'SECS';
+
+  return {
+    name: exercise.name,
+    color: exercise.color,
+    type: exerciseType,
+    sets: exercise.sets.map((set) => ({
+      prevWeight: set.weight && set.weight.toString(),
+      weight: '',
+      prevDuration: exerciseType === 'REPS' ? set.reps.toString() : set.time.toString(),
```
ctperry0301, line 36: The set.reps.toString() and set.time.toString() will cause an error if they are left blank

components/CreateCycle/ExpandableWorkoutCard.jsx
```diff
@@ -76,12 +76,12 @@ const ExpandableWorkoutCard = (props) => {
         <FlatList
           data={exercises}
           // TODO: rename to exercise name
```
ctperry0301: can these TODO's be completed now?

ctperry0301 (commented):
Yeah, we will definitely need a loading page. 
See comments, close to approval

components/LogWorkout/LogWorkout.jsx
jpoist97: Yeah, they should never both be undefined though. They will be for newly created workouts but once AddWorkout is updated to reflect the new workout structure it should never occur

components/CreateCycle/ExpandableWorkoutCard.jsx
jpoist97: Yeah, I'll take care of it

components/CreateCycle/ExpandableWorkoutCard.jsx
```diff
@@ -74,14 +89,13 @@ const ExpandableWorkoutCard = (props) => {
...
         <FlatList
-          data={exercises}
-          // TODO: rename to exercise name
+          data={parsedExercises}
           keyExtractor={(item) => item.id}
           renderItem={({ item }) => (
             <RowContent>
-              <WorkoutViewText style={{ marginLeft: 10 }}>{item.id}</WorkoutViewText>
+              <WorkoutViewText style={{ marginLeft: 10 }}>{item.name}</WorkoutViewText>
               {/* TODO: center the sets x reps around the x so double digit numbers dont look bad */}
```
jpoist97, line 97: Aye @finlaylp do you remember what this todo was referring to?

finlaylp: yeah that was so that like 3 x 12 and 3 x 8 would be aligned around the x instead of just centered, bc it looks a bit odd that the x's can sometimes be offset a bit. Not a big issue but I didn't get around to it.

ctperry0301 (approved):
Loads much faster now, Looks good.

PR #47, opened 2021-01-11
Auth final

finlaylp:
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-11 at 10 41 07
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-11 at 10 41 19
Added forgot password to the login page. It's kind of mostly handled by firebase


components/Authentication/login.jsx
```diff
@@ -60,36 +59,55 @@ const ShowText = styled.Text`
...
       });
   }
 
   function signupPress() {
     navigation.navigate('Signup');
   }
 
-  firebase.auth().onAuthStateChanged(() => {
-    if (firebase.auth().currentUser) {
-      navigation.navigate('Root');
```
jpoist97, line 90: Do we need this to auto-login if the user has logged in before?

components/Authentication/signup.jsx
```diff
@@ -60,46 +59,90 @@ const ShowText = styled.Text`
...
+    setValidEmail(!ValidateEmail(email));
+  }, [email]);
+
+  React.useEffect(() => {
+    if (isFirstRunPassword.current) {
+      isFirstRunPassword.current = false;
+      return;
+    }
+    setValidPassword(!(password.length > 5));
+  }, [password]);
```
jpoist97: I don't quite understand this code, I didn't see your comment last night and idk where it went anymore haha.

jpoist97 (commented):
Same general comments as last time, forgot password looks good! One small thing that would be nice is if you tapped off of the text the keyboard would hide, but not 100% necessary rn

components/Authentication/login.jsx
finlaylp: no. that's done in app.js see line 136

components/Authentication/signup.jsx
finlaylp: All good. It is basically just there to stop it from triggering the first time, since useEffect triggers on init of a state var so I didn't want all the invalid field text things to show when the user hasn't even clicked on the input lines. So after they click on the input, it will allow it to show the invalid text thing.

finlaylp:
> Same general comments as last time, forgot password looks good! One small thing that would be nice is if you tapped off of the text the keyboard would hide, but not 100% necessary rn

I'll try and add that rq

components/Authentication/login.jsx
jpoist97: Does it still work? It doesn't do it for me anymore when I run it locally

finlaylp: hmm. Well this is just a listener for auth state change, so if they log in or log out it triggers. So it shouldn't be what is automatically logging you in

components/Authentication/signup.jsx
```diff
@@ -60,107 +53,155 @@ const ShowText = styled.Text`
...
+      {!validName && <ViewFiller />}
+      <InputLine
+        placeholder="Email"
+        selectionColor="#A192FF"
+        textContentType="emailAddress"
+        value={email}
+        onChangeText={(text) => setEmail(text)}
+      />
+      {validEmail && <View style={{ paddingLeft: 15, width: '80%', flexDirection: 'row' }}><ErrorText>Invalid email.</ErrorText></View>}
+      {!validEmail && <ViewFiller />}
```
jpoist97: nit: These can be simplified into one statement with a ternary operator.

components/Authentication/signup.jsx
jpoist97: Okay, that makes sense. I feel like there's probably an easier way to do this, but I can't think of one off the top of my head and I think its fine cause this works

jpoist97 (commented):
Looks good to me, only concern is the auto-logging in stuff, otherwise I think this should be ready to go in soon.

components/Authentication/signup.jsx
finlaylp: I thought so too. This was the only way I could figure out, it was more of an issue than I expected

components/Authentication/signup.jsx
finlaylp: good catch

components/Authentication/login.jsx
jpoist97: hm does it still work for you? Maybe it takes a second for Firebase to realize you're logged in so this was catching it previously?

finlaylp: Not working for me either, although I'm not sure if it should be remembering you between simulator runs since it is a new package on expo. In theory App.js should be doing that part, after initializing firebase it checks for a user. By default firebase keeps users logged in so that should not be null if they are. But idk I'll look into it more before merging

jpoist97: Okay, I think everything else in the PR is ready to merge so once we figure that out you can just re-request my review and I'll approve it

ctperry0301 (dismissed):
In addition to what JP said, make sure that you dont leave any TODOs in this. Otherwise, this is really good. Good stuff

App.js
```diff
@@ -133,12 +133,15 @@ const App = () => {
...
   let startupScreen = 'Login';
-  if (user) {
-    startupScreen = 'Root';
-    const displayName = user.displayName;
-  }
+
+  firebase.auth().onAuthStateChanged((user) => {
+    //user is already logged in
+    if (user) {
+      const displayName = user.displayName;
```
jpoist97: Is this used anywhere?

components/Authentication/login.jsx
```diff
@@ -60,91 +53,123 @@ const ShowText = styled.Text`
...
+    }
+  }
+
   firebase.auth().onAuthStateChanged(() => {
-    if (firebase.auth().currentUser) {
+    if(firebase.auth().currentUser) {
       navigation.navigate('Root');
     }
-  });
+  })
```
jpoist97: Do we still need this here?

jpoist97 (dismissed):
Looks good, just comments that don't affect functionality

App.js
jpoist97: referring to the displayName variable

components/Authentication/login.jsx
finlaylp: Nope accident good catch

App.js
finlaylp: well it will be used as the welcomeName once we update that from being 'Shriya'. But I can delete it since rn it's unnecessary

components/Authentication/signup.jsx
```diff
@@ -60,107 +53,149 @@ const ShowText = styled.Text`
...
+    }
+    return (false);
+  }
+
+  React.useEffect(() => {
+    if (isFirstRunName.current) {
+      isFirstRunName.current = false;
+      return;
+    }
+    setValidName(!((name.length > 1)));
```
ctperry0301, line 92: if you put in else around each of these "actions", then you could combine the three useEffectes into one while still having the same functionality.

ctperry0301: In this case tho, I think it's better to have multiple useEffects like you have. Ignore the comment above.

jpoist97 (approved):
Looks ready for me, as long as Cole's cool with it I think this can go in

components/Authentication/signup.jsx
finlaylp: yeah cuz they're only triggering when the specific field they correlate with is changed, whereas if they were all in one they'd all trigger from a single of the textinputs

PR #48, opened 2021-01-16
Data sent to database on workout & cycle creation

ctperry0301:
For testing, make sure currentUser is as follows in home.jsx:
      // const currentUser = firebase.auth().currentUser.uid;
      const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';

### DIFF
--- a/components/CreateWorkout/CreateWorkout.jsx
+++ b/components/CreateWorkout/CreateWorkout.jsx
@@ -89,19 +89,18 @@ export default ({ navigation }) => {
     setItemState(exerciseList);
   }
 
-  const sendWorkoutToDB = (createWorkout) => {
+  const sendWorkoutToDB = (newWorkout) => {
     const currentUser = firebase.auth().currentUser.uid;
-    //const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';
 
     const dbRef = firebase.firestore();
     const userRef = dbRef.collection('users').doc(currentUser);
     const workoutRef = userRef.collection('workouts');
 
-    createWorkout = JSON.parse(JSON.stringify(createWorkout, (k, v) => {
+    newWorkout = JSON.parse(JSON.stringify(newWorkout, (k, v) => {
       if (v === undefined) { return null; } return v;
     })); // This is needed so that we can have an undefined weight and color
 
-    workoutRef.add(createWorkout);
+    workoutRef.add(newWorkout);
   };
 
   return (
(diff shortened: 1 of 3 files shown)


components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -4,10 +4,12 @@ import * as React from 'react';
 import { Alert, View } from 'react-native';
 import styled from 'styled-components/native';
 import { useDispatch, useSelector } from 'react-redux';
+import * as firebase from 'firebase';
 import FinishButton from '../utils/FinishButton';
 import PlusButton from '../utils/PlusButton';
 import AdjustExercisesList from './AdjustExercisesList';
 import { ADD_WORKOUT } from '../../constants';
+// import * as admin from 'firebase-admin'
```
jpoist97:
```suggestion

```

components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -88,6 +90,21 @@ export default ({ navigation }) => {
     setItemState(exerciseList);
   }
 
+  const sendWorkoutToDB = (newWorkout) => {
+    // const currentUser = firebase.auth().currentUser.uid;
+    const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';
```
jpoist97: Should we swap what is commented here?

jpoist97 (approved):
Nice looks good

components/CreateWorkout/CreateWorkout.jsx
shriyan44: yes

components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -88,6 +90,21 @@ export default ({ navigation }) => {
...
+
+    const dbRef = firebase.firestore();
+    const userRef = dbRef.collection('users').doc(currentUser);
+    const workoutRef = userRef.collection('workouts');
+
+    newWorkout = JSON.parse(JSON.stringify(newWorkout, (k, v) => {
+      if (v === undefined) { return null; } return v;
+    })); // This is needed so that we can have an undefined weight and color
+
+    workoutRef.add(newWorkout);
```
jpoist97, line 103: Wait jk, I think that we might need to format the data a little bit when we send it off. For example, we need the time/reps instead of duration. We also want that to be updated in the store as well

jpoist97 (changes requested):
Yeah, can you format the workout to have sets with time or reps. Also make sure this updated workout object is what is in the store since this is what's causing it to error now.

components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -129,7 +144,40 @@ export default ({ navigation }) => {
...
+              for (let i = 0; i < sets; i++) {
+                setArr.push({ 
+                  weight: undefined,
+                  time: item.seconds || '10'
+                });}
+              }
+              return {
+                sets: setArr,
+                exerciseId: item.id,
+              };
```
jpoist97: I think you can do this with another map on the sets array

components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -129,7 +144,40 @@ export default ({ navigation }) => {
               };
             }),
           };
-
+          let muscleGroups = itemState[0].muscleGroups;
```
jpoist97, line 126: Do we need two different objects for redux and database? I think they can we the same right?

jpoist97 (commented):
Looks good, see comments but should go in soon

jpoist97 (commented):
Changes look good. One thing I forgot to mention is the workouts in redux need their ids as an additional field in the workout. If you could just add that in this would be done!

jpoist97 (approved):

PR #49, opened 2021-01-16
Signup db pop

finlaylp:
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-18 at 12 53 37
Colors don't seem to be working but all of the data is entered correctly when a user signs up


package.json
```diff
@@ -24,6 +24,7 @@
     "expo-status-bar": "~1.0.2",
     "expo-updates": "~0.3.3",
     "firebase": "7.9.0",
+    "firebase-admin": "^9.4.2",
```
jpoist97: Is this being used anywhere?

components/Authentication/signup.jsx
```diff
@@ -115,7 +118,151 @@ const Signup = ({ navigation }) => {
...
+              }).catch((error) => {
+                console.log(error);
+              });
+
+            userRef.update({selectedCycleId: cyclesDataRef.id})
+              .then(() => {
+                console.log("selected cycle updated.")
+              }).catch((error) => {
+                console.log(error);
+              });
```
jpoist97: Could we do these in parallel as well?

components/Authentication/signup.jsx
```diff
@@ -115,7 +118,151 @@ const Signup = ({ navigation }) => {
...
+            Promise.all(
+              userDataRef,
+              userWorkout1.set(pushWorkout), 
+              userWorkout2.set(pullWorkout),
+              userWorkout3.set(legsWorkout)
+            ).then(() => {
+              console.log("workouts and user set successfully.");
+            }).catch((error) => {
+              console.log(error);
+            });
```
jpoist97: Can we make this an async function and use await on these promises and the other promises. The reason why I say this is right now we're navigating to home **before** Firebase is done setting up a user's information. 

Not sure if this is the issue but I hope it is cause its an easy fix. But when I create a new user and it logs me in right away, my app is empty and my firebase responses are empty as well. I'm guessing its because the info isn't set in firebase yet so we need to wait for it to finish setting up and then navigate to home.

jpoist97 (changes requested):
Looking good, see comment about async/await and should be good to go in assuming that fixes the bug

package.json
finlaylp: No I'll get rid of it

components/Authentication/signup.jsx
finlaylp: In parallel with each other or in parallel with all of the other ones?

jpoist97: At least with each other, optimally it'd be with the other stuff but I'm not sure if we can do those together

finlaylp: Right I think I have that working in my updated code, I don't think it's possible to get them to all run in parallel tho

jpoist97 (approved):
Code looks good to go in, one thing we should add in the future is something on the sign up button that shows we're loading in between signing up and navigating to root

PR #50, opened 2021-01-17
Connect home to db

rohithdara:
- displays current user's name at top of home page
- workout list in home page shows 5 most recent workouts (displays all if less than 5)


jpoist97 (approved):
Looks good

PR #51, opened 2021-01-18
workout log to db progress

shriyan44:


components/Home/Home.jsx
```diff
@@ -101,8 +101,8 @@ export default () => {
   useEffect(() => {
     const initializeDatabase = async () => {
       // Get the current logged in user id
-      const currentUser = firebase.auth().currentUser.uid;
-      // const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';
+      // const currentUser = firebase.auth().currentUser.uid;
```
rohithdara: should switch back which line is commented before merging to main

components/LogWorkout/LogWorkout.jsx
```diff
@@ -77,6 +80,59 @@ const LogWorkout = (props) => {
 
   const [exerciseState, setExerciseState] = useState(initialExerciseState);
 
+
+const updatePrevDetails = (props) => exerciseState.map((exercise) => {
+  const exerciseType = exercise.sets[0].reps ? 'REPS' : 'SECS';
+  return {
+    ID: exercise.exerciseId,
+    name: 'AYOOOOOO',
```
rohithdara: Is this temporary or why is the name AYOOOOOO? I'm confused lol

rohithdara (dismissed):
I tried hopping on your branch and logging the push workout. I remember you saying that the "undefined is not an object (evaluating 'set.time.toString')" was only broken on the arms one, but I got the same error for push since bench press set to time. Have you been able to find a fix? 
Other than that, everything else looks good!

components/CreateCycle/CreateCycle.test.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import EllipsisPopup from '../utils/EllipsisPopup';
+
+it('renders correctly', () => { 
+    const tree = renderer
+    .create(<EllipsisPopup></EllipsisPopup>).toJSON();
+    expect(tree).toMatchSnapshot();
+});
```
jpoist97: Did you mean to include this in the pr?

components/LogWorkout/LogWorkout.jsx
```diff
@@ -27,6 +30,7 @@ const parseExercises = (exercises) => exercises.map((exercise) => {
   const exerciseType = exercise.sets[0].reps ? 'REPS' : 'SECS';
 
   return {
+    ID: exercise.exerciseId,
```
jpoist97:
```suggestion
    id: exercise.exerciseId,
```

Can we make this consistent with other code.

components/LogWorkout/LogWorkout.jsx
```diff
@@ -77,6 +81,56 @@ const LogWorkout = (props) => {
...
+                }
+                _.set(parsedSet, exercise.type === 'REPS' ? ['reps'] : ['time'], set.duration || set.prevDuration)
+                return parsedSet;
+              })
+          }
+
+      })
+  
+    }
+    userRef.collection('workouts').doc(workoutId).update(newWorkoutLog);
```
jpoist97: I don't think this currently updates lastPerformed, could you make sure that gets updated as well

components/LogWorkout/LogWorkout.jsx
```diff
@@ -167,8 +223,13 @@ LogWorkout.propTypes = {
...
     params: PropTypes.shape({
       workoutId: PropTypes.string.isRequired,
+      isSelectedCycle: PropTypes.bool,
     }),
   }).isRequired,
 };
 
+LogWorkout.defaultProps = {
+  isSelectedCycle: false,
+};
```
jpoist97, line 243: You need to put the route, params stuff outside for this to work as expected

components/LogWorkout/LogWorkout.jsx
```diff
@@ -77,6 +81,56 @@ const LogWorkout = (props) => {
...
+    // userRef.update({selectedCycleIndex: 2})
+    // userRef.get().then(function(doc) {
+    //       console.log("Document data:", doc.data().selectedCycleIndex);
+    //   } 
+    // )
+    return (
+    <CurrentCycle
+          rightPress={() => { dispatch({ type: INCREMENT_SELECTED_CYCLE_INDEX, cycleLength: cycleDetails.length }); }}
+        /> )
+  };
```
jpoist97: I don't think this does what you want it to do. You just need a function that calls dispatch to increment selected cycle index and also updates the database as well. The stuff ou have commented out on line 91 is pretty close for the database part

jpoist97 (commented):
See comments, close to being done

components/CreateCycle/CreateCycle.test.js
shriyan44: teehee

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +85,49 @@ const LogWorkout = (props) => {
...
+  // const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';
+
+  const dbRef = firebase.firestore();
+  const userRef = dbRef.collection('users').doc(currentUser);
+  const dispatch = useDispatch();
+
+  const incrementSelectedCycleIdx = () => {
+    userRef.update({ selectedCycleIndex });
+    dispatch({ type: INCREMENT_SELECTED_CYCLE_INDEX, cycleLength });
+  };
```
jpoist97: What's the userRef.update doing here? To me this looks like it's not updating the selectedCycleIndex properly because you're sending it to firebase before incrementing it. Also if you want to access selectedCycleIndex you don't need to pass it through props, you can just access it directly from redux

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +85,49 @@ const LogWorkout = (props) => {
...
+      date: firebase.firestore.FieldValue.serverTimestamp(),
+      exercises: exerciseState.map((exercise) => ({
+        exerciseId: exercise.id,
+        exerciseName: exercise.name,
+        sets: exercise.sets.map((set) => {
+          const parsedSet = {
+            weight: set.weight || set.prevWeight,
+          };
+          _.set(parsedSet, exercise.type === 'REPS' ? ['reps'] : ['time'], set.duration || set.prevDuration);
+          return parsedSet;
```
jpoist97: nice

reducers/workoutReducer.js
```diff
@@ -14,6 +14,9 @@ const workoutReducer = (state = initialState, action) => {
       case ADD_WORKOUT:
          console.log(`Adding workout to store with ${action.workout}`);
          const newWorkouts = [...state.workouts];
+         newWorkouts.map((workoutEle) => {
+            console.log(workoutEle.id)
```
jpoist97: Delete this

reducers/workoutReducer.js
```diff
@@ -24,6 +27,24 @@ const workoutReducer = (state = initialState, action) => {
...
+                  lastPerformed: 0,
+               };
+            } 
+            else {
+               return workoutEle;
+            }
+         });
+         return {
+            workouts: newW
+         }
```
jpoist97: Can we rename newW to something better

actions/workoutActions.js
```diff
@@ -21,8 +21,16 @@ const deleteWorkout = (workoutId) => {
    }
 }
 
+const updateWorkoutPrev = (workout) => {
+   return {
+      type: UPDATE_WORKOUT_PREV,
+      workout,
+   }
+}
```
jpoist97: Can we change this to just take in **workoutId and updatedExercises**. Those are the only two fields you use from the workout and it'll be confusing later since we'll think this requires an entire workout object

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +85,49 @@ const LogWorkout = (props) => {
...
+      })),
+    };
+
+    const newWorkoutDoc = {
+      workoutId: newWorkoutLog.workoutId,
+      lastPerformed: newWorkoutLog.date,
+      exercises: newWorkoutLog.exercises,
+    };
+
+    userRef.collection('workouts').doc(workoutId).update(newWorkoutDoc); // updates workout doc
```
jpoist97, line 127: This code stores extra information in firebase, we don't want the workoutId to be stored here

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +85,49 @@ const LogWorkout = (props) => {
...
+  const incrementSelectedCycleIdx = () => {
+    userRef.update({ selectedCycleIndex });
+    dispatch({ type: INCREMENT_SELECTED_CYCLE_INDEX, cycleLength });
+  };
+
+  const sendWorkoutLogToDB = () => {
+    const workoutRecsRef = userRef.collection('workoutRecords');
+
+    const newWorkoutLog = {
+      workoutName: name,
```
jpoist97, line 106: Workout logs do not need name

components/LogWorkout/LogWorkout.jsx
jpoist97: Yeah I tested this and the selectedCycleIndex is not incremented in firebase, it'll work if you access it from redux put line 96 below the dispatch

jpoist97 (commented):
Looking good, see comments for things to fix

jpoist97:
Also general point, did we say we didn't want people to be able to log a workout unless everything was completed

components/LogWorkout/LogWorkout.jsx
jpoist97: Just kidding actually, I think we should store the workout name and exercise name in the workout log so Rohith and Fin don't need to cross-reference them in the calendar component

components/LogWorkout/LogWorkout.jsx
```diff
@@ -99,12 +99,23 @@ const LogWorkout = (props) => {
...
+      const formatMonth = (month) => {
+        if (month < 10) {
+          return '0' + month;
+        }
+        else {
+          return month;
+        }
+      }
+      return date.getFullYear() + '-' + formatMonth(date.getMonth() + 1) + '-' + date.getDate();
+    }
```
jpoist97: Use date-fns to format the date rather than writing it yourself

jpoist97 (commented):
Code looks good, I think you're just missing the alert when someone tries to log workout and they haven't completed everything yet

components/LogWorkout/LogWorkout.jsx
rohithdara: lol I wrote that, but I sent a date-fns version to replace this with

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +86,49 @@ const LogWorkout = (props) => {
...
+  const dispatch = useDispatch();
+
+  const incrementSelectedCycleIdx = () => {
+    dispatch({ type: INCREMENT_SELECTED_CYCLE_INDEX, cycleLength });
+    userRef.update({ selectedCycleIndex: (cycleIdx + 1) % cycleLength });
+  };
+
+  const sendWorkoutLogToDB = () => {
+    const workoutRecsRef = userRef.collection('workoutRecords');
+    const { format } = require('date-fns');
```
jpoist97: Can you change this to an import at the top of the code

components/LogWorkout/LogWorkout.jsx
```diff
@@ -146,7 +199,19 @@ const LogWorkout = (props) => {
...
       <StyledFinishButton onPress={() => {
         // TODO: Update Redux and database
-        navigation.goBack();
+        const completedList = [];
+        exerciseState.forEach((exercise) => {
+          exercise.sets.forEach((set) => {
+            completedList.push(set.completed);
+          });
+        });
+        if (completedList.includes(false)) {
```
jpoist97: I don't think this is the best way to do this. I think we could replace this with nested for loops that exit on the first false they find. That way if the first one is false we don't waste time on the rest of them and we don't need to call another .includes after to check

jpoist97: Also can you take out the TODO now that you've implemented that functionality

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +86,49 @@ const LogWorkout = (props) => {
...
+  const currentUser = firebase.auth().currentUser.uid;
+  // const currentUser = '68w6wWz8l5QJO3tDukh1fRXWYjD2';
+
+  const dbRef = firebase.firestore();
+  const userRef = dbRef.collection('users').doc(currentUser);
+  const dispatch = useDispatch();
+
+  const incrementSelectedCycleIdx = () => {
+    dispatch({ type: INCREMENT_SELECTED_CYCLE_INDEX, cycleLength });
+    userRef.update({ selectedCycleIndex: (cycleIdx + 1) % cycleLength });
```
jpoist97: Can you flip these two lines. I know it works this way but I'm a bit worried that since you're using the actual selectedCycleId this might increment it 2 times in the database (even though it doesn't). It's just weird to me that it's not being re-rendered after your action completes.

jpoist97 (commented):
Look good! I'm excited to use the app once this goes in. One other thing is did Fin and Rohith want you to also add to like a list of dates in the user when they log a workout?

rohithdara:
Fin and I have that list working in our PR, so no need for Shriya to do it

jpoist97:
Also this shouldn't post any data to firestore as "n/a", it should be null if the value is "n/a"

components/LogWorkout/LogWorkout.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import * as firebase from 'firebase';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
+import _, { set } from 'lodash';
 import styled from 'styled-components';
```
jpoist97: Remove { set } and test this again. I think this is gonna be used instead of the actual set inside your map

components/LogWorkout/LogWorkout.jsx
```diff
@@ -111,9 +112,9 @@ const LogWorkout = (props) => {
         exerciseName: exercise.name,
         sets: exercise.sets.map((set) => {
           const parsedSet = {
-            weight: set.weight || set.prevWeight,
+            weight: (set.prevWeight == 'n/a' && set.weight  == '') ? null : parseInt(set.weight) || parseInt(set.prevWeight),
```
jpoist97: Use triple equals in javascript please.

Would something like this work better:
`parseInt(set.weight) || set.prevWeight === 'n/a' ? null : parseInt(set.prevWeight)`

This way we only check that prevWeight is n/a when we have to. Also I'm p sure NaN is falsey so you don't need to worry about set.weight being an empty string.

jpoist97 (commented):
Also please address my old comments and fix the build

shriyan44:
> Look good! I'm excited to use the app once this goes in. One other thing is did Fin and Rohith want you to also add to like a list of dates in the user when they log a workout?

  just modified the format that the date was being sent in idk bout no list

components/LogWorkout/LogWorkout.jsx
shriyan44: ya lol idk why I thought oof the logic like that haha

components/LogWorkout/LogWorkout.jsx
shriyan44: sorry dont understand what u want

components/LogWorkout/LogWorkout.jsx
shriyan44: I get what u mean and that was kinda my first way of doing it so I had like an if false then show alert and return, but I was confused on the logic cuz I didnt know how to do it so that the finish functions only get called if all are completed

components/LogWorkout/LogWorkout.jsx
jpoist97: instead of `const { format } = require('date-fns')` here in the middle of your code move it to the top and import it like all of the other modules used in this file. `import { format } from 'date-fns'`

components/LogWorkout/LogWorkout.jsx
jpoist97: There's a number of ways you can do it. Here's one off the top of my head

```
let completed = true;
for(let i = 0; completed && i < exerciseState.length; i++) {
   const sets = exerciseState[i].exercise.sets;
   for(let j = 0; completed && j < sets.length; j++) {
      if(!set[j].completed) {
         completed = false;
      }
   }
}

if(!completed) {
   <Show alert>
} else {
   <Do database stuff>
}

```

It might not work exactly correctly but the logic is there

jpoist97:
> > Look good! I'm excited to use the app once this goes in. One other thing is did Fin and Rohith want you to also add to like a list of dates in the user when they log a workout?
> 
> just modified the format that the date was being sent in idk bout no list

Rohith said they did that in their pr so no need for you to do it

components/LogWorkout/LogWorkout.jsx
shriyan44: wow I rlly forgot about flags that was so incredibly simple my b lmao

components/LogWorkout/LogWorkout.jsx
shriyan44: actually I tried your way and it doesnt work it was sending null when I set a weight so I just changed it back to mine

jpoist97 (dismissed):
Done, just need to fix merge conflicts

jpoist97 (approved):

PR #52, opened 2021-01-22
added button disabling for auth and ran lint

finlaylp:
Nothing big, just added button disabling.

### DIFF
--- a/components/Authentication/login.jsx
+++ b/components/Authentication/login.jsx
@@ -159,7 +159,6 @@ const Login = ({ navigation }) => {
           <LoginButton
             uppercase={false}
             mode="contained"
-            dark
             onPress={loginPress}
           >
             <LoginText>Login</LoginText>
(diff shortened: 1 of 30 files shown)


components/Authentication/login.jsx
```diff
@@ -147,14 +154,27 @@ const Login = ({ navigation }) => {
...
+            uppercase={false}
+            mode="contained"
+            dark
+            onPress={loginPress}
+          >
+            <LoginText>Login</LoginText>
+          </LoginButton>
+        )
+        : (
+          <DisableLoginButton
```
ctperry0301, line 167: Would u want this button to have a disabled={disableButton} like you do the sign up one?

ctperry0301 (commented):
Im pretty sure there is a better way to conditionally render a style than to use a ternary for the whole button. The change works great and looks good tho.

components/Authentication/login.jsx
finlaylp: yeah good catch

finlaylp:
> Im pretty sure there is a better way to conditionally render a style than to use a ternary for the whole button. The change works great and looks good tho.

Yeah there must be I just couldn't figure it out and this seemed like an easy fix anyway

ctperry0301 (approved):
Looks Good.

PR #53, opened 2021-01-22
Implement alternating colors on frontend

jpoist97:


components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -123,30 +123,30 @@ export default ({ navigation }) => {
...
+            muscleGroups += ' ';
             muscleGroups += itemState[1].muscleGroups;
           }
           const newWorkout = {
             name,
             lastPerformed: 'n/a',
             id: nextWorkoutId,
-            muscleGroups: muscleGroups,
+            muscleGroups,
             color: itemState[0].color,
```
ctperry0301: If kept, the color would be stored in the database which defeats the purpose of rendering it on the front end. (Is this needed for front end rendering or redux or something?) If not, then for this task, I think you should also get rid of all instances relating color to redux and db.

ctperry0301 (changes requested):
The front end rendering part works great. However, I think there is still some code lingering from when we thought we were gonna store colors in the database. Clean out that code and I think it should be good to go.

components/CreateWorkout/CreateWorkout.jsx
jpoist97: I removed those instances when I fixed the cycle bug, they should all be gone now

ctperry0301 (approved):
Works Great, good stuff!

PR #54, opened 2021-01-22
Fix Cycle bug and update selected cycle in firebase

jpoist97:
Fixed bug where cycles would not show up after reloading app. Also added functionality so that you can not select a cycle and it's updated in firebase.

### DIFF
--- a/components/Home/Home.jsx
+++ b/components/Home/Home.jsx
@@ -188,6 +188,7 @@ export default () => {
           leftPress={() => { dispatch(actions.cycles.decrementSelectedCycleIndex(cycleDetails.length)); }}
           rightPress={() => { dispatch(actions.cycles.incrementSelectedCycleIndex(cycleDetails.length)); }}
           id={cycleDetails && selectedCycle.workouts[cycles.selectedCycleIndex]}
+          isCycleSelected={cycleDetails !== undefined}
         />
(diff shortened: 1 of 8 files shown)


components/Home/Home.jsx
```diff
@@ -128,37 +128,36 @@ export default () => {
...
       }
     });
-    return workoutList.slice(0,5);
-  }
-
+    return workoutList.slice(0, 5);
+  };
 
   const selectedCycle = (cycles.selectedCycleId !== undefined) && _.find(cycles.cycles, (cycle) => cycle.id === cycles.selectedCycleId);
   let cycleDetails;
```
ctperry0301: You will get an index undefined error if you try to click on a left or right arrow on the home page if a cycle isn't set

ctperry0301: > You will get an index undefined error if you try to click on a left or right arrow on the home page if a cycle isn't set

Im also getting the same error if I try to create a cycle while not having a cycle currently selected (new account).

ctperry0301 (changes requested):
I dont think this code is working as expected for me, The cycles arent getting loaded in for me when I try and test it.

jpoist97:
Thanks for catching that Cole, I forgot to test making a new account. All the bugs should be fixed now

ctperry0301 (approved):
Good Stuff!

PR #55, opened 2021-01-22
Fix all linting

ctperry0301:
A looooooot of files were changed, Although I tested each aspect thoroughly, it should betested further before approval. 

check eslintrc.js file and make sure that were okay with the ignored rules

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,11 +18,11 @@ module.exports = {
     'react',
   ],
   rules: {
-    'max-len': 'off',
-    'radix': 'off',
-    'global-require': 'off',
-    'no-param-reassign': 'off',
-    'react/forbid-prop-types': 'off',
-    'react/prop-types': 'off',
+    'max-len': 'off', // Kind of useless, for code style
+    'radix': 'off', // Useless, we wont be dealing with hex values
+    'global-require': 'off', // this one is fine because were requiring image filepaths,  
+    'react/forbid-prop-types': 'off', // this ideally would be fixed, but were in too deep 
+    'react/prop-types': 'off', // this ideally would be fixed, but were in too deep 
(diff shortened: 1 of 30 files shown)


ctperry0301:
Also, this should probably be merged in first so that other PR's wont have linting issues

components/CreateCycle/ExerciesView.jsx
```diff
@@ -0,0 +1,86 @@
+import { useSelector } from 'react-redux';
+import { View } from 'react-native';
+import { FlatList } from 'react-native-gesture-handler';
+import styled from 'styled-components/native';
+import React from 'react';
+import _ from 'lodash';
```
jpoist97, line 6: Thank you for moving this out

.eslintrc.js
```diff
@@ -18,5 +18,11 @@ module.exports = {
     'react',
   ],
   rules: {
+    'max-len': 'off', // Kind of useless, for code style
+    'radix': 'off', // Useless, we wont be dealing with hex values
+    'global-require': 'off', // this one is fine because were requiring image filepaths,  
+    'react/forbid-prop-types': 'off', // this ideally would be fixed, but were in too deep 
+    'react/prop-types': 'off', // this ideally would be fixed, but were in too deep 
+    'no-unused-expressions': ["error", { "allowTernary": true }] // Although ternary isn't optimized, it is still useful
```
jpoist97: I think we should leave this on. I took a look at all of the places where this rule gets violated and those should all be changed to if/else statements.

jpoist97 (commented):
Thanks for doing all of this Cole, I just have one comment

.eslintrc.js
ctperry0301: Yeah, good call. Should be fixed now

jpoist97 (approved):
Looks good to me. Thanks Cole you the homie for this one

PR #58, opened 2021-01-28
Implemented Delete Workout Functionality

jpoist97:
Implemented delete workout functionality. Users cannot delete workouts from the selected cycle. Also added thunk so we can combine our actions and database calls.

I split delete and edit workouts into different tasks so the review process is easier


api.js
```diff
@@ -0,0 +1,91 @@
+import * as firebase from 'firebase';
+import 'firebase/firestore';
+
+/**
+ * Helper function that returns a reference to the current logged in user in 
+ * Firestore.
+ */
+const getUserRef = () => {
```
finlaylp, line 8: if u made userRef a param for the functions u could only have to make the call once

components/Workouts/AlphabetWorkoutList.jsx
```diff
@@ -83,7 +84,13 @@ const AlphabetWorkoutList = (props) => {
         name={left.name}
         subtext={left.subtext}
         displayEllipses={left.displayEllipses}
-        deleteWorkout={() => dispatch(actions.workouts.deleteWorkout(left.id))}
+        deleteWorkout={() => {
+          if (selectedCycle.workouts.includes(left.id)) {
+            alert('This workout is included in the selected cycle. Please selected another cycle before deleting this workout.');
```
finlaylp: tiny grammar - select instead of selected in the second sentence

components/Workouts/AlphabetWorkoutList.jsx
```diff
@@ -94,7 +101,13 @@ const AlphabetWorkoutList = (props) => {
           name={right.name}
           subtext={right.subtext}
           displayEllipses={right.displayEllipses}
-          deleteWorkout={() => dispatch(actions.workouts.deleteWorkout(right.id))}
+          deleteWorkout={() => {
+            if (selectedCycle.workouts.includes(right.id)) {
+              alert('This workout is included in the selected cycle. Please selected another cycle before deleting this workout.');
```
finlaylp: same here

finlaylp (dismissed):
code looks good

api.js
jpoist97: Yeah you're right. I think I'm gonna leave ti this way so there's not like an extra step when you want to call an api function. I think even though this setup calls getUserRef more than is needed, its not a big deal bc its not that expensive to do so

components/Workouts/AlphabetWorkoutList.jsx
jpoist97: oops hahah good catch

finlaylp (approved):
Looks good! I was thinking the same thing with the userRef, I imagine retrieving the user is about as inexpensive as it gets with firebase

PR #59, opened 2021-01-30
Calendar component

finlaylp:
The only thing missing here is that the current day on bootup doesn't start as selected, you have to click on it to get the day's workout records to show. when you click on a day it puts it in redux after pulling from firestore so any clicks after that load the workout record instantly
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-30 at 15 46 44
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-30 at 15 46 58
Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-30 at 15 46 51


rohithdara:
Damn looks fire 😍🔥

Gustavotd2 (changes requested):
M

components/Calendar/Calendar.jsx
```diff
@@ -1,10 +1,166 @@
...
+import styled from 'styled-components';
+import firebase from 'firebase';
+import { useDispatch, useSelector } from 'react-redux';
+import CalendarWorkoutCard from './CalendarWorkoutCard';
+import actions from '../../actions';
+
+const days = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'];
+const months = ['January', 'February', 'March', 'April', 'May', 'June', 'July',
+  'August', 'September', 'October', 'November', 'December'];
+
```
jpoist97: Can you move these to the constants file

reducers/pastWorkoutDatesReducer.js
```diff
@@ -0,0 +1,26 @@
...
+            return {
+                dates: [...action.dates]
+            };
+        case ADD_WORKOUT_RECORD_DATE:
+            console.log(`Adding workout to store with ${action.date}`);
+            const newDates = [...state.dates];
+            newDates.push(action.date);
+            return {
+                dates: newDates,
+            };
```
jpoist97: Should we check for duplicate dates here

reducers/workoutRecordReducer.js
```diff
@@ -0,0 +1,21 @@
...
+
+const recordReducer = (state = initialState, action) => {
+   switch(action.type) {
+        case ADD_WORKOUT_RECORD:
+            console.log(`Adding workout to store with ${action.record}`);
+            const newRecords = [...state.records];
+            newRecords.push(action.record)
+            return {
+                records: newRecords,
+            };
```
jpoist97, line 15: Did you guys consider storing the records in an object like a hashmap with the dates as the key and the record as the value. Cause we're not gonna show a list of the records anywhere right so this will make it faster to retrieve

jpoist97: This isn't a huge deal though and we can just fix this going forward

components/LogWorkout/LogWorkout.jsx
```diff
@@ -76,6 +87,48 @@ const LogWorkout = (props) => {
...
+
+    const newWorkoutDoc = {
+      lastPerformed: newWorkoutLog.date,
+      exercises: newWorkoutLog.exercises,
+    };
+
+    userRef.collection('workouts').doc(workoutId).update(newWorkoutDoc); // updates workout doc
+    dispatch(actions.workouts.updateWorkoutPrev(workoutId, newWorkoutLog.exercises)); // rerenders workout to show update prev details
+    workoutRecsRef.add(newWorkoutLog); // makes a new workoutRecord
+  };
```
jpoist97: Is the code to put the new dates in Firebase in here?

components/Calendar/Calendar.jsx
```diff
@@ -1,10 +1,166 @@
...
+  const marks = {};
+  dates.forEach((date) => {
+    marks[date] = { marked: true, selectedColor: '#cab0ff' };
+  });
+  return marks;
+}
+
+const CalendarView = () => {
+  const startDate = new Date();
+  const stateStart = `${days[startDate.getDay()]}, ${months[startDate.getMonth()]} ${startDate.getDate()}`;
```
jpoist97: Is there not a function in date-fns that can do this for you?

components/Calendar/Calendar.jsx
```diff
@@ -1,10 +1,166 @@
...
+
+const NoWorkoutText = styled(DayTitle)`
+  font-size: 24px;
+`;
+
+function formatDate(date) {
+  return `${days[date.getDay()]}, ${months[date.getMonth()]} ${date.getDate() + 1}`;
+}
+
+function setDates(dates) {
```
jpoist97: Can we rename this function to something more descriptive. Maybe markCalendarDates or something like that

components/Calendar/Calendar.jsx
```diff
@@ -1,10 +1,166 @@
...
+
+        const workoutSets = [];
+
+        exerciseRecord.forEach((record) => {
+          const loggableData = {
+            name: record.exerciseName,
+            sets: record.sets,
+            isReps: true,
+            date,
+          };
```
jpoist97: Does this support time based exercises?

components/Calendar/Calendar.jsx
```diff
@@ -1,10 +1,166 @@
...
+          if (allDates[date.dateString]) {
+            allDates[date.dateString] = { selected: true, marked: allDates[date.dateString].marked, selectedColor: '#cab0ff' };
+          } else {
+            allDates[date.dateString] = { selected: true, marked: false, selectedColor: '#cab0ff' };
+          }
+          setMarkedDates(allDates);
+
+          if (firstRun.current) {
+            firstRun.current = false;
+          }
```
jpoist97, line 124: What is firstRun used for?

jpoist97 (commented):
This looks great! Good job guys this is super cool.

A couple of small things that came up when I was testing it by had. 
1) If you log a workout and then go to the calendar it will not pull from Firebase so the day just says no workouts performed.
2) When I did that the weight came up as null lbs. And the workoutName was not displayed either.
3) Can we make the colors of the cards alternate, you should be able to just reuse the logic used for all the other cards.
4) The build is failing so you'll have to run the linting again before we can merge this pr.

IMG_4103

components/Calendar/Calendar.jsx
finlaylp: I took a brief look through date-fns and couldn't really get exactly what I wanted so I thought this'd just be easier

components/Calendar/Calendar.jsx
finlaylp: the isReps field is actually redundant bc time vs reps is determined in the flatlist on line 148 so I'll prob delete it

components/Calendar/Calendar.jsx
finlaylp: when the calendar component first loads up it says "no date selected". I think we were planning on changing that to having the current day automatically selected but forgot to implement it

reducers/workoutRecordReducer.js
finlaylp: yeah I was thinking about that didn't get around to it yet

reducers/pastWorkoutDatesReducer.js
finlaylp: yep good catch

components/LogWorkout/LogWorkout.jsx
finlaylp: @shriyan44 @rohithdara not sure I didn't do this part I don't think

components/Calendar/Calendar.jsx
finlaylp: Oh we would also still probably need it if we wanted to have today's date selected when it loads up

components/LogWorkout/LogWorkout.jsx
rohithdara: I don't know what happened within the commits and pulls, but the call to the function that updated it was lost. I added it back in with the most recent commit.

components/.expo/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "hostType": "lan",
+  "lanType": "ip",
+  "dev": true,
+  "minify": false,
+  "urlRandomness": null,
+  "https": false
```
jpoist97: Can we take these out if the're not doing anything

jpoist97 (commented):
I think this can go in as is honestly, just take out those expo settings files. I think that this component can be improved, but we can refine it in the next sprint or later down the line

components/Workouts/.expo/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "hostType": "lan",
+  "lanType": "ip",
+  "dev": true,
+  "minify": false,
+  "urlRandomness": null,
+  "https": false
+}
```
jpoist97: Delete this

jpoist97 (approved):

PR #60, opened 2021-01-31
Implemented edit workouts

jpoist97:
Implemented edit workouts and updated create workout code to no longer add duplicate muscle groups


api.js
```diff
@@ -91,6 +91,12 @@ export async function deleteCycleDocument(cycleId) {
 }
 
 
+export async function updateWorkoutDocument(workoutId, newWorkoutContent) {
+   const userRef = getUserRef();
```
finlaylp, line 95: if functions keep requiring userRef, at some point it might help to just make it a param so u only call once at the start of the app. But I doubt it matters much

components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -37,9 +38,60 @@ const AddCycleButton = styled(PlusButton)`
...
+  // Get two unique muscleGroups
+  const allMuscleGroups = itemState.map((item) => item.muscleGroups);
+  const workoutMuscleGroups = _.uniq(allMuscleGroups);
+  const topMuscleGroups = workoutMuscleGroups.slice(0, 2).join(' ');
+
+  return {
+    muscleGroups: topMuscleGroups,
+    exercises: itemState.map((item) => {
+      const setArr = [];
+      const sets = item.sets || 3;
```
ctperry0301, line 53: dont let item.sets be 0

components/CreateWorkout/CreateWorkout.jsx
```diff
@@ -37,9 +38,60 @@ const AddCycleButton = styled(PlusButton)`
...
+      muscleGroups: matchingExercise.muscleGroups,
+      sets: exercise.sets.length.toString(),
+      reps,
+      seconds,
+      isReps: !!reps,
+    };
+  };
+
+  const [name, setName] = React.useState(_.get(editWorkout, ['name'], ''));
+  const [itemState, setItemState] = React.useState(editWorkout ? editWorkout.exercises.map(parseWorkoutExercises) : []);
```
finlaylp, line 94: should the map be done outside of the initial state? Won't it try and map every time this component re renders

finlaylp (dismissed):
Looks great

components/CreateWorkout/CreateWorkout.jsx
jpoist97: Nope, actually the initial state code only runs the very first time the component is created, so this code will only map once

api.js
jpoist97: I definitely see what you're saying but I don't think requiring it as a param is the way to go. That'd just mean we'd have to grab the userRef everywhere we call this function from. I guess we could store it in redux, but idk if it's really worth it

ctperry0301 (approved):
Looks good! Great stuff

api.js
finlaylp: true prob not worth it

jpoist97: Honestly, it's actually not a half bad idea

PR #61, opened 2021-02-01
Added Edit/Delete Cycle Functionality

ctperry0301:


components/CreateCycle/CreateCycle.jsx
```diff
@@ -50,20 +53,22 @@ export default ({ navigation }) => {
...
+    } else {
+      cycleDoc = await cycleRef.doc(cycleID).update(newCycle);
+    }
 
     dispatch(actions.cycles.addCycle({
       workouts: newCycle.workoutIds,
       name: newCycle.name,
       id: cycleDoc.id,
     }));
-    navigation.goBack();
```
jpoist97, line 66: Is there a reason you decided to move this out of this function. This causes it to abruptly add the cycle in after like half a second if you create a new cycle

components/Cycles/Cycles.jsx
```diff
@@ -19,18 +19,37 @@ export default () => {
...
         <AlphabetCycleList items={allCycles} selectedCycle={selectedCycle} />
-        <AddCycleButton title="Cycle" size={18} onPress={() => navigation.navigate('Create Cycle')} />
+        <AddCycleButton
+          title="Cycle"
+          size={18}
+          onPress={() => navigation.navigate('Create Cycle', {
+            cycleName: '',
+            cycleDetails: [],
+            isNewCycle: true,
+          })}
```
jpoist97, line 50: Not that it really matters but you could do this with default PropTypes in CreateCycle component

components/CreateCycle/CreateCycle.jsx
```diff
@@ -50,20 +53,22 @@ export default ({ navigation }) => {
...
+      cycleDoc = await cycleRef.add(newCycle);
+    } else {
+      cycleDoc = await cycleRef.doc(cycleID).update(newCycle);
+    }
 
     dispatch(actions.cycles.addCycle({
       workouts: newCycle.workoutIds,
       name: newCycle.name,
       id: cycleDoc.id,
     }));
```
jpoist97: Does this code add a new cycle to redux even when editing?

jpoist97: This actually breaks when you edit a workout. It just shows as a warning bc its a promise rejection but when I edit a cycle I get `[Unhandled promise rejection: TypeError: undefined is not an object (evaluating 'cycleDoc.id')]` I think you can fix this by creating a new action/reducer for specifically updating a cycle and call that when editing instead of addCycle

jpoist97 (changes requested):
When I was testing I noticed a couple issues with color. When you edit a cycle all of the workout cards are pink, I think you'll just have to parse them to add color if they're editing a cycle.

When I was running the code it didn't work for me and I got that promise rejection error when trying to edit. lmk if you need more help figuring out redux.

jpoist97:
Oh also if you're trying to run your code and it doesn't work anymore you'll have to merge from master bc the exercises have different format now

components/CreateCycle/CreateCycle.jsx
```diff
@@ -50,20 +53,27 @@ export default ({ navigation }) => {
...
+    } else {
+      await cycleRef.doc(cycleID).update(newCycle);
+      dispatch(actions.cycles.updateCycle(cycleID,
+        {
+          id: cycleID,
+          workouts: newCycle.workoutIds,
+          name: newCycle.name,
+        }));
+    }
+    navigation.navigate('Cycles');
```
jpoist97: does navigaion.goBack() not work here? I think we should use that instead if it works here

jpoist97 (dismissed):
One really minor comment, looks good to me!

components/CreateCycle/CreateCycle.jsx
ctperry0301: Fixed

jpoist97 (approved):

PR #62, opened 2021-02-01
Changed muscleGroups to string

jpoist97:
Changed all exercises to have muscleGroups as a string instead of array. Left the code I ran commented out in api.js for future reference


rohithdara (approved):
looks good, out of curiosity -- how/where in the file did you actually run the function

shriyan44:
Looks good, I see the updated muscle group in firestore too

shriyan44 (approved):

jpoist97:
I just called it from home, you could call it anywhere it doesn't matter. But don't run it again bc it'll mess up the muscleGroups

PR #63, opened 2021-02-06
Sort by muscle group

rohithdara:
- add exercise bug is fixed
- can toggle between sorting by muscle group and name
- searching now returns an alphabetical flatlist instead of an alphabet section list
- can now search any part of the word and it'll match instead of just being able to match with the beginning of the word
- only design issue is the scroll tab covering the + icons... Regardless you can click anywhere on the card (not just the +) for it to select properly

Screen Shot 2021-02-05 at 4 01 45 PM
Screen Shot 2021-02-05 at 4 01 49 PM
Screen Shot 2021-02-05 at 4 02 11 PM

### DIFF
--- a/components/AddExercises/AddExercises.jsx
+++ b/components/AddExercises/AddExercises.jsx
@@ -52,6 +52,7 @@ export default ({ route }) => {
         onExercisesAdd={route.params.onExercisesAdd}
         parsedItemsName={parseItemsByName(exerciseObjects)}
         parsedItemsMuscleGroups={parseItemsByMuscleGroup(exerciseObjects)}
+        exerciseObjects = {exerciseObjects}
       />
     </ModalScreenWrapper>
   );
--- a/components/AddExercises/AddExercises.jsx
+++ b/components/AddExercises/AddExercises.jsx
@@ -5,16 +5,16 @@ import ModalScreenWrapper from '../utils/ModalScreenWrapper';
 
 export default ({ route }) => {
   const exercises = useSelector((state) => state.exercises.exercises);
-  const items = exercises.map((exercise) => ({ ...exercise, subtext: exercise.muscleGroups }));
+  const exerciseObjects = exercises.map((exercise) => ({ ...exercise, subtext: exercise.muscleGroups }));
 
   const parseItemsByName = (items) => {
     // Sort names alphabetically
     items.sort((a, b) => a.name.localeCompare(b.name));
-  
+
     // Group by first letter of each name
     const bucketData = items.reduce((accumulator, item) => {
       const bucket = item.name[0].toUpperCase();
-  
+
       // If this is the first time we've seen this letter, create a bucket
       if (!accumulator[bucket]) {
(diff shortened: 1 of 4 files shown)


components/AddExercises/AddExercises.jsx
```diff
@@ -5,11 +5,54 @@ import ModalScreenWrapper from '../utils/ModalScreenWrapper';
...
+      bucketData[key].sort((a, b) => a.name.localeCompare(b.name));
+    });
+    return bucketData;
+  };
 
   return (
     <ModalScreenWrapper>
-      <ExerciseList items={items} onExercisesAdd={route.params.onExercisesAdd} />
+      <ExerciseList
+        onExercisesAdd={route.params.onExercisesAdd}
```
jpoist97, line 53: nit: can we add this to props validation or use _.get to make sure we don't error out if route is undefined

components/AddExercises/ExerciseList.jsx
```diff
@@ -36,92 +39,111 @@ const Buttontext = styled.Text`
...
+            }
+          });
+        }
+        filteredItems.sort((a, b) => a.name.localeCompare(b.name));
+      });
+      setDataState({ ...dataState, filteredDataSource: filteredItems });
+      setSearch(text);
+    } else {
+      setDataState({ ...dataState, filteredDataSource: dataState.masterDataSource });
+      setSearch(text);
```
jpoist97: Consider passing in the original exerciseList and then just filtering that list so you don't have to deal with this as an object

components/AddExercises/ExerciseList.jsx
```diff
@@ -36,92 +39,111 @@ const Buttontext = styled.Text`
...
+      if (allExercises[item.name[0]][index].selected === true) {
+        setExerciseCount(exerciseCount + 1);
+        addedExercises.push(item);
+      } else {
+        setExerciseCount(exerciseCount - 1);
+        addedExercises.splice(addedExercises.indexOf(item), 1);
+      }
     }
+    setDataState({ ...dataState, masterDataSource: allExercises });
+  };
```
jpoist97: I think we can get rid of the if/else block here. If we just conditionally save the first array key in a variable (item.muscleGroups or item.name[0]) you can reuse all the code inside the if/else block.

jpoist97 (commented):
Looks really good, only code style comments

jpoist97 (dismissed):

components/AddExercises/ExerciseList.jsx
```diff
@@ -188,5 +178,20 @@ const ExerciseList = ({
     </SafeAreaView>
   );
 };
+ExerciseList.propTypes = {
+  route: PropTypes.shape({
+    params: PropTypes.shape({
+      onExercisesAdd: PropTypes.func.isRequired,
+    }),
+  }).isRequired,
```
jpoist97: Should go in addExercises

jpoist97 (approved):

PR #66, opened 2021-02-17
Create custom exercises

ctperry0301:
Added Ability to Create Custom Exercises.


actions/exerciseActions.js
```diff
@@ -7,6 +7,15 @@ const initalizeExercises = (exercises) => {
    }
 }
 
+const addCustomExercise = (exercises) => {
+  return {
+     type: ADD_CUSTOM_EXERCISE,
+     exercises,
```
jpoist97: nit: should just be exercise right?

components/AddExercises/CreateCustomExercise.jsx
```diff
@@ -0,0 +1,171 @@
...
+import RNPickerSelect from 'react-native-picker-select';
+import * as firebase from 'firebase';
+import actions from '../../actions/index';
+
+const Name = styled.Text`
+font-family: 'Montserrat_600SemiBold';
+font-size: 16px;
+margin-left: 8px;
+margin-bottom: 10px;
+padding: 4px;
```
jpoist97: Style: Could you tab these over a once, they don't get picked up by eslint for some reason

components/AddExercises/CreateCustomExercise.jsx
```diff
@@ -0,0 +1,171 @@
...
+    const customExerciseRef = firebase.firestore()
+      .collection('users')
+      .doc(currentUser)
+      .collection('customExercises');
+    const ExerciseDoc = await customExerciseRef.add(exercise);
+    dispatch(actions.exercises.addCustomExercise({
+      muscleGroups: muscleGroup,
+      name: workoutName,
+      id: ExerciseDoc.id,
+    }));
```
jpoist97: Could you move most of this code into an action that uses thunk, you can pretty much just copy this code over. Also could you add the creating exercise doc as a function in api.js

jpoist97 (commented):
Mostly nitpicks, make sure I correctly fixed the merge conflicts and this should be good to go in.

components/AddExercises/CreateCustomExercise.jsx
ctperry0301: Im not quite sure how to do this, is there a good, easy, example somewhere?

jpoist97: Yeah I think a good example is in workoutActions, the updateWorkout function.

Basically you just have the action return a function that takes in dispatch as the only parameter. From there you can do whatever you want in the function and then dispatch an action to the reducer. So all you'd do in your action creator is call the database stuff from api and then dispatch the action that your current actionCreator already dispatches

jpoist97 (approved):

PR #67, opened 2021-02-17
Loading screen

shriyan44:


components/Home/LoadingScreen.jsx
```diff
@@ -0,0 +1,134 @@
...
+      exercises.push({
+        id: doc.id,
+        name,
+        muscleGroups,
+      });
+    }
+  });
+
+  return exercises;
+};
```
jpoist97, line 78: Can probably wait until next sprint but we should move these to thunk actions and functions in api.js

components/Home/Home.jsx
```diff
@@ -173,15 +65,25 @@ export default () => {
...
+          top: 20,
         }}
       />
-      <View style={{ marginBottom: '10%', marginTop: '5%' }}>
-        <WelcomeTitle>Hello</WelcomeTitle>
-        <WelcomeTitle numberOfLines={1}>
-          {welcomeName}
-          !
-        </WelcomeTitle>
+      <View style={{ marginBottom: '10%', marginTop: '5%', marginLeft: '5%' }}>
```
rohithdara, line 75: I think we made the standard to style the components at the top and not inline, not 100% sure if required though

components/Home/Home.jsx
```diff
@@ -173,15 +65,25 @@ export default () => {
...
+
+        </Animatable.Text>
+        <Animatable.Text animation="bounceIn" iterationCount={1}>
+          <WelcomeTitle numberOfLines={1}>
+            {welcomeName}
+            !
+          </WelcomeTitle>
+        </Animatable.Text>
       </View>
       <View style={{ height: '50%', marginBottom: '25%' }}>
```
rohithdara, line 88: same here

rohithdara (commented):
looks fine to me, just some styling tings with using styled components rather than inline styles

oh also the bounce animation for the name is almost unnoticeable, I think it starts too quickly (or maybe even during) the switch between loading screen and home so you don't see the start of it

finlaylp (approved):
good stuff

jpoist97 (approved):

PR #69, opened 2021-02-18
Implement thunk action for incrementing/decrementing selectedCycleIndex

jpoist97:
Pretty simple, just made it so now the database updates on the arrow presses and consolidated into a thunk action


ctperry0301 (approved):
Looks good

PR #70, opened 2021-02-18
Refined progress page

jpoist97:
Removed duplicate actions, fixed when exercise name too long in chart, update exerciseRecord store on logWorkout, fixed unnecessary eslint disable


components/Profile/ExerciseChart.jsx
```diff
@@ -90,7 +89,7 @@ export default () => {
   return (
     <ExerciseChartContainer>
       <TwoColumnView>
-        <ExerciseName>{exerciseState.selectedExerciseName || 'Select an Exercise'}</ExerciseName>
+        <ExerciseName longText={exerciseState.selectedExerciseName && exerciseState.selectedExerciseName.length >= 20}>{exerciseState.selectedExerciseName || 'Select an Exercise'}</ExerciseName>
```
ctperry0301, line 92: This fix should work most of the time, but in the case where they create a custom exercise with a really long name, we could run into the same issue. Should we cap the custom exercise name length at a certain amount of letters to prevent this from happening.

ctperry0301 (dismissed):
See 1 small comment, but overall looks really good!

components/Profile/ExerciseChart.jsx
jpoist97: Added 25 character limit to custom exercise names

ctperry0301 (approved):

PR #72, opened 2021-02-19
Progress bar

rohithdara:
Added to the top of log workout.
- Updates properly even if sets are added/deleted
- Animated, but I couldn't figure out gradients as color was the components own property, not a style choice
Screen Shot 2021-02-19 at 12 00 59 PM
Screen Shot 2021-02-19 at 12 01 03 PM
Screen Shot 2021-02-19 at 12 01 14 PM


finlaylp (approved):
Looks good

PR #74, opened 2021-02-20
basic settings modal

shriyan44:
extras:
- deleted fakedata file
- set signup password to show default


components/Profile/SettingsPopup.jsx
```diff
@@ -1,54 +1,135 @@
...
+// };
 
-SettingsPopup.defaultProps = {
-  style: [],
-  triggerSize: 18,
-};
+// SettingsPopup.defaultProps = {
+//   style: [],
+//   triggerSize: 18,
+// };
```
jpoist97: I think you need to put back in triggerSize if you're gonna use it

components/Profile/SettingsPopup.jsx
```diff
@@ -1,54 +1,135 @@
...
+import React, { useState } from 'react';
 import {
-  Menu, MenuOptions, MenuOption, MenuTrigger,
-} from 'react-native-popup-menu';
-import Proptypes from 'prop-types';
+  Alert, Modal, StyleSheet, Pressable, View, TouchableOpacity, Image,
+} from 'react-native';
 import styled from 'styled-components/native';
 import { Ionicons } from '@expo/vector-icons';
+import firebase from 'firebase';
```
jpoist97: Can we rename this file to SettingsModal.jsx

components/Profile/SettingsPopup.jsx
```diff
@@ -1,54 +1,135 @@
...
+    justifyContent: 'center',
+    alignItems: 'center',
+    marginTop: 10,
+  },
+  modalView: {
+    height: 500,
+    width: 350,
+    margin: 10,
+    backgroundColor: 'white',
+    borderRadius: 20,
```
jpoist97: This should be done using styled components

components/Profile/SettingsPopup.jsx
```diff
@@ -1,54 +1,135 @@
...
-          backgroundColor: '#EDEDF0', borderRadius: 10, padding: 5, width: 'auto',
-        },
-      }}
+    <View>
+      <Modal
+        animationType="slide"
+        transparent
+        visible={modalVisible}
+        onRequestClose={() => {
+          Alert.alert('Modal has been closed.');
```
jpoist97: What is this for?

components/Profile/SettingsPopup.jsx
```diff
@@ -1,54 +1,135 @@
...
-              <StyledText>{option.text}</StyledText>
-            </Text>
-          </MenuOption>
-        ))}
-
-      </MenuOptions>
-    </Menu>
+
+        <TouchableOpacity style={styles.centeredView} onPress={() => setModalVisible(!modalVisible)}>
+          <TouchableOpacity style={styles.modalView} onPress={() => setModalVisible(modalVisible)}>
```
jpoist97: Why are these touchableOpacities? Do Views not work for these and why do you need the onPress stuff?

components/Profile/SettingsPopup.jsx
```diff
@@ -1,54 +1,135 @@
...
+            <TouchableOpacity style={styles.colorButton} onPress={() => { alert('clicked blue'); }}>
+              <Image source={require('../../assets/blue.png')} />
+            </TouchableOpacity>
+            <TouchableOpacity style={styles.colorButton} onPress={() => { alert('clicked purple'); }}>
+              <Image source={require('../../assets/purple.png')} />
+            </TouchableOpacity>
+            <Pressable
+              style={[styles.button]}
+              onPress={async () => {
+                setModalVisible(false);
```
jpoist97: I think we should do this after they are successfully logged out

jpoist97 (dismissed):
This is pretty much done, I left some questions I had but honestly I can just fix those things when I add to the modal cause I wanna start working on it soon. Another question I had was what's the difference between the Pressable and TouchableOpacity

components/Profile/SettingsPopup.jsx
shriyan44: I had to use touchable opacity and on presses cuz its like the only workaround for making modal close when u touch out of it

finlaylp (approved):

shriyan44:
> This is pretty much done, I left some questions I had but honestly I can just fix those things when I add to the modal cause I wanna start working on it soon. Another question I had was what's the difference between the Pressable and TouchableOpacity

theyre basically same but perusable has more optional props like a prop that defines how far a touch can register away from the the wrapped element so ill prolly use that for the small buttons

PR #75, opened 2021-02-22
Rest notifications

jpoist97:
Implemented rest notifications, notifications will de disabled by default and can be enabled in settings. They are scheduled once you click completed on any set besides the last set of an exercise.

IMG_4156
IMG_4157
IMG_4158


components/Profile/SettingsModal.jsx
```diff
@@ -78,12 +109,47 @@ const SettingsModal = (props) => {
...
+                    ...settingState,
+                    enableRestNotifications: !settingState.enableRestNotifications,
+                  });
+                }}
+                value={settingState.enableRestNotifications}
+              />
+            </TwinView>
+
+            <TwinView>
+              <SettingLabel disabled={!settingState.enableRestNotifications}>Rest Notification Timer:</SettingLabel>
```
rohithdara: Can you add units to the time (like seconds or minutes)

rohithdara: Also, if you enable notifications, then clear the input field and save changes -- it'll show as NaN

rohithdara: I'm pretty sure this also causes a "Malformed calls from JS: field sizes are different" error when you click done on a set.

rohithdara: Ok also, if you fill out the field for the time and don't save changes, it still looks like it changed on the front end but doesn't actually do anything in the back-end which can be confusing

rohithdara (changes requested):
So I couldn't actually get the notifications to show up for me on the ios simulator, but that is probably the simulator's fault. Made some UI/UX comments on the rest timer on the settings page.

jpoist97:
Yeah @rohithdara I forgot to mention in my original PR that the notifications only show up on actual devices, they won't show up on the simulator.

components/Profile/SettingsModal.jsx
```diff
@@ -59,31 +61,115 @@ const ModalView = styled(TouchableOpacity)`
...
+  border-radius: 5px;
 `;
 
 const SettingsModal = (props) => {
   const { triggerSize } = props;
   const [modalVisible, setModalVisible] = useState(false);
   const navigation = useNavigation();
+  const dispatch = useDispatch();
+  const settings = useSelector((state) => state.settings);
+  const [settingState, setSettingState] = useState({
```
finlaylp, line 101: This is kinda nit-picky, but if you took out the async part in the redux action I think it would simplify this a bit because it would remove the need for the local settings state as well as redux settings. Then maybe when the settings modal is closed it could update to the user profile in the db in the background. Just an idea, this looks like it works fine as is.

finlaylp (commented):
I couldn't get it to open on my phone (wifi issue not code) so couldn't test notifs, so I'll wait for someone else to approve this PR, but the code looks good. One thing -- do we not have to worry about asking for permissions when the app first boots up? I think that iOS requires apps to ask for notification permission so that the actual settings in your iphone get edited. I briefly read about it [here](https://docs.expo.io/versions/latest/sdk/permissions/) but not sure if we need them.

components/Profile/SettingsModal.jsx
jpoist97: > I couldn't get it to open on my phone (wifi issue not code) so couldn't test notifs, so I'll wait for someone else to approve this PR, but the code looks good. One thing -- do we not have to worry about asking for permissions when the app first boots up? I think that iOS requires apps to ask for notification permission so that the actual settings in your iphone get edited. I briefly read about it [here](https://docs.expo.io/versions/latest/sdk/permissions/) but not sure if we need them.

Yeah you're right I totally forgot abt that let me add it in rn

jpoist97:
Actually now the notifications should work on the simulator if you enable them, it will only ask the first time for you, otherwise if you click don't allow we cannot request again and the user must go into the settings themselves and enable them

PR #78, opened 2021-02-24
Remove workout from cycle

ctperry0301:


components/CreateCycle/CreateCycle.jsx
```diff
@@ -45,12 +45,30 @@ export default ({ navigation, route }) => {
...
+      if (workouts[i].name === workoutName) {
+        workouts.splice(i, 1);
+      }
+    }
+    if (!isNewCycle) {
+      // Makes sure selectedCycleIndex decreases if number of workouts in cycle decreases
+      dispatch(actions.cycles.decrementSelectedCycleIndex(cycles.selectedCycleIndex, cycleDetails.length));
+    }
+    updateOrder(workouts);
+  }
```
jpoist97: Couple comments on this function:
1) I think you could simplify this function a bit if you just took in the index of the workout instead of the workoutName, you wouldn't need to do the for loop in the middle
2) Also, correct me if I'm wrong, but this code would decrement the selected cycle index if the cycle is being edited, not just if it was the selected cycle.

components/CreateCycle/DraggableWorkoutList.jsx
```diff
@@ -18,6 +18,7 @@ const DraggableWorkoutList = (props) => {
   const renderCard = ({ item, index, drag }) => (
     <ExpandableWorkoutCard
       drag={drag}
+      removeWorkout={removeWorkout}
```
jpoist97: If you swap removeWorkout to take in the index of the workout you can pass it here, you'll need to use currying but I can help you out with that if you need help

jpoist97 (changes requested):
Couple of comments about optimizing the removing and the selectedCycle stuff

jpoist97 (approved):
Looks good, nice currying

PR #81, opened 2021-02-25
Moved loading screen database operations to actions

jpoist97:
Moved loading screen db ops to actions. Added loading to store as well as comments to api.js file.


ctperry0301 (approved):
Works like it should! Thanks for doing this!

PR #82, opened 2021-02-25
Add colors

finlaylp:
Way too many files changed to add pictures here, but it's fully working so u can check out the PR to see what they look like. 
I know the calendar part is bad code but there was honestly no other way to do it, I spent like 3 hours just trying to get the calendar marks to update but I don't think it is possible.


components/Home/Home.jsx
```diff
@@ -102,7 +103,7 @@ export default () => {
         <CurrentCycle
           name={cycleDetails && cycleDetails[cycles.selectedCycleIndex].name}
           subtext={cycleDetails && cycleDetails[cycles.selectedCycleIndex].muscleGroups}
-          color={cycleDetails && COLORS[cycles.selectedCycleIndex % COLORS.length]}
+          color={cycleDetails && COLORS[colorTheme][cycles.selectedCycleIndex % (COLORS[colorTheme].length - 1)]}
```
rohithdara, line 106: I think the < and > for the selected cycle are part of the purple color scheme, might be a good idea to change that color too because it feels a little out of place

components/Profile/ExerciseChart.jsx
```diff
@@ -103,7 +104,7 @@ export default () => {
       { loading
         ? (
           <SpinnerContainer>
-            <ActivityIndicator color={COLORS[0]} size="large" />
+            <ActivityIndicator color={COLORS[colorTheme][0]} size="large" />
```
rohithdara, line 107: It's not letting me comment on the actual button, but clicking the "Select Exercises" causes an invalid hook error.

components/CreateCycle/ExerciesView.jsx
```diff
@@ -69,11 +69,10 @@ const ExercisesView = (props) => {
       </RowHeader>
       <FlatList
         data={parsedExercises}
-        keyExtractor={(item) => item.id}
+        keyExtractor={(item, index) => item.id + index.toString()}
```
rohithdara, line 72: I know you didn't make it but can you change it to ExercisesView lol

rohithdara (changes requested):
Overall good looks, main fix is the 'Select Exercises' button in the profile tab, and a couple other stylistic things. Also, with this change, we need to make alternative icons as well right because it looks odd with the old icon on the new color scheme. Codewise looks like the same change everywhere so I didn't really have any comments for that haha

rohithdara (approved):
Good work, now we just gotta implement the new icon once those are made.

PR #86, opened 2021-02-26
added new logo

finlaylp:
Can't put anything dynamic in require so this is the easiest way to do it


constants/index.js
```diff
@@ -45,6 +45,10 @@ export const COLORS = {
   'default': ['#CAB0FF', '#9D8DFF', '#6D8DFF', '#4457BC'],
   'aqua': ['#4EC9DA', '#65C6BE', '#7DD8FF', '#86DDB4'],
 };
+export const LOGOS = {
+  'default' : require('../../assets/logo.png'),
```
ctperry0301: I think it should just be one ../ for these routes

finlaylp: No cuz this is the path relative to Home.jsx, not index.js

jpoist97: I think Cole is right here, when I try to run the app as is it gives an error, but when I switch it to just one ../ it runs as expected

finlaylp: yeah he was right forgot to come back and update it

rohithdara (approved):

ctperry0301 (approved):
looks good

PR #87, opened 2021-02-27
fixed bug that showed n/a days ago

rohithdara:
This is literally a 1 line PR. Someone toss it a review, it's just to make sure that 'Try First Time' works in an edge case. I added an extra condition because strings return as false if you check if it is NaN.


jpoist97:
your changes looked good to me @rohithdara , when I was testing it I noticed that there was another bug with the lastPerformed which caused it to always be NaN, I just added the fix to this branch cause they're related and it's hella small

jpoist97 (approved):

PR #90, opened 2021-03-03
Finishing Actions Tests (Settings, WorkoutRecord, Workout)

rohithdara:
settingsActions and workoutRecordActions are at 100%
workoutActions is at 60%

The other 40% is fairly complicated so I'm leaving it as-is for now, unless we want the full 100% with everything.

### DIFF
--- a/__tests__/actionCreators/settingsActions.test.js
+++ b/__tests__/actionCreators/settingsActions.test.js
@@ -1,7 +1,16 @@
 import settingsActions from '../../actions/settingsActions';
+import * as api from '../../api';
 import { UPDATE_USER_SETTINGS } from '../../constants/index';
 
+jest.mock('../../api.js', () => ({
+   updateUserSettings: jest.fn()
+}));
+
 const mockDispatch = jest.fn();
+
+afterEach(() => {
+   jest.clearAllMocks();
+});
 
 describe('settingsActions tests', () => {
     describe('initialize settings tests', () => {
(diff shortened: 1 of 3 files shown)


__tests__/actionCreators/settingsActions.test.js
```diff
@@ -0,0 +1,38 @@
+import settingsActions from '../../actions/settingsActions';
+import { UPDATE_USER_SETTINGS } from '../../constants/index';
+
+const mockDispatch = jest.fn();
+
+describe('settingsActions tests', () => {
```
jpoist97: Small thing but make sure you add the afterEach code here that resets the mockDispatch function. It's int he workoutAction tests but missing here

jpoist97: Also I think you need to mock updateUserSettings in this file as well since it's called from updateSettings.

jpoist97 (changes requested):
Looking good, couple comments and it'll be ready to go in. Just make sure after you mock the one function from api in settingsActions tests that api.js is not listed in the coverage report since we shouldn't have any tests for it.

__tests__/actionCreators/settingsActions.test.js
rohithdara: sir yes sir

jpoist97 (approved):
Looks good!

PR #94, opened 2021-03-05
Complete cycleReducer tests

jpoist97:


rohithdara (approved):
Well, I went through all of it the best I could. It runs, passes, and meets coverage so looks good to me!

PR #96, opened 2021-03-05
Final fixes

finlaylp:
should fix the bugs on sonar cloud

### DIFF
--- a/components/Calendar/CalendarWorkoutCard.jsx
+++ b/components/Calendar/CalendarWorkoutCard.jsx
@@ -85,11 +85,9 @@ const CalendarWorkoutCard = (props) => {
       <FlatList
         data={sets}
         keyExtractor={(item, index) => item.exerciseID + index.toString()}
-        renderItem={({ item, index }) => {
-          return (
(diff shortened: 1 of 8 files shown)


components/Calendar/CalendarWorkoutCard.jsx
```diff
@@ -66,7 +66,7 @@ const CalendarWorkoutCard = (props) => {
 
   SetRow.propTypes = {
     index: PropTypes.number.isRequired,
-    lbs: PropTypes.number,
+    lbs: PropTypes.any,
```
jpoist97, line 69: Just curious why did we need to make this change?

jpoist97 (commented):
Changes look good, just one small question and then this can go in

components/Calendar/CalendarWorkoutCard.jsx
finlaylp: To allow 'N/A' as a value so that it doesn't say null lbs

jpoist97 (approved):

PR #97, opened 2021-03-06
Workout tests

finlaylp:
Screen Shot 2021-03-05 at 8 23 48 PM


reducers/workoutRecordReducer.js
```diff
@@ -17,6 +17,7 @@ const recordReducer = (state = initialState, action) => {
                 prevRecords[date] = [record];
             }
             return {
+                ...state,
```
jpoist97, line 20: nice catch

jpoist97 (approved):
I didn't look too hard at the tests but they pass so looks good to me

PR #103, opened 2021-08-21
[TS Conversion][Phase 1] Convert Reducers/Actions/Constants to .ts

jpoist97:
Converted remaining files to `.ts` files.


rohithdara (approved):
I mean I see no reason this wouldn't be approved haha. But can you talk a little bit on what the process looks like to make this conversion

jpoist97:
Yeah I just forgot that I opened this PR, the new CI thing I switched us to takes a lot longer than Travis did.

But in terms of what is needed to make the switch to TS I'm following the 3 step process in this youtube video. This PR wraps up phase 1, and the next step is to start removing our implicit any typing. here's the [link to the video](https://www.youtube.com/watch?v=OdU9H-_d14Y&list=WL&index=26&t=344s&ab_channel=CodingTech) and step 2 starts at about 5:47.

